PR #22, opened 2015-12-09
Replace express with native node calls

gregbty:
I updated the old express calls to use the ones provided by the base node framework. 

There are two other changes I wanted to suggest as well. If a module is not found an error is thrown which results in a non-graceful shutdown of the process. This in turn results in a generic exception in .NET which does not give any indication of the error that actually occurred within the node process.

```node
throw new Error('The module "' + resolvedPath + '" has no export named "' + bodyJson.exportedFunctionName + '"');
``` 

This also applies to the logic used in `react-rendering.js`. I would propose ensuring that all errors bubble up to the client without causing the process to end unexpectedly i.e. wrapping the logic inside of a try/catch block and sending the error via the http pipeline before ending the process.

I can submit an additional PR if needed.

### DIFF
--- a/Microsoft.AspNet.NodeServices/Content/Node/entrypoint-http.js
+++ b/Microsoft.AspNet.NodeServices/Content/Node/entrypoint-http.js
@@ -20,7 +20,8 @@ var server = http.createServer(function(req, res) {
             if (!hasSentResult) {
                 hasSentResult = true;
                 if (errorValue) {
-                    res.status(500).send(errorValue);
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for this! It's merged.

Regarding your other suggestion about error handling, I just added commit 80575a that hopefully addresses this. Let me know if you think anything further is needed.

gregbty:
That should work. Thanks!

Issue #31, opened 2016-01-17
Update to use Babel 6


xabikos:
This is an awesome project and I founded recently. I am very interested especially for React and server side rendering part of it. I am trying to improve the tooling on the field so I have published a [nuget package](https://www.nuget.org/packages/Webpack/) [GitHub project](https://github.com/xabikos/aspnet-webpack) which simplifies the webpack integration. Especially for simple scenarios like the sample an external configuration file is not required and all the references to the scripts are added automatically. I tried to add it to the sample and make a pull request but when update the babel to latest versions (6+) I am getting an error for server side rendering. Are there any plans to update to the latest babel?

SteveSandersonMS:
Yes, absolutely! I'll take a look at your pull request shortly. Thanks for your contribution!

PR #32, opened 2016-01-17
Upgrade to Babel 6 and update the ReactGrid sample

xabikos:
After creating the #31 issue I did a bit of investigation in the code and found is not a big effort to start using Babel 6. Please let me know if the pull request is good enough and if there are further changes that should take place


SteveSandersonMS:
Merged - thanks!

PR #47, opened 2016-04-11
Add sudo recommendation for dnu restore

MarkPieszak:


SteveSandersonMS:
Could you clarify why we should recommend this? I'm primarily working on a Mac and I never use `sudo` with `dnu restore`. Why would you need to? It seems like that would open an unnecessary security risk.

SteveSandersonMS:
Oh I didn't actually mean to close this, but now I have done, I'll leave it closed unless you think we really should do this (my guess is that it's probably not necessary). Sorry if it seemed rude to close it so abruptly!

MarkPieszak:
@SteveSandersonMS 
Personally on 2 MACs I work on I ned to use `sudo` ahead of a lot of `npm installs` / etc, maybe I could update the wording to say:

 (On Mac you *may* need to prefix this with `sudo`) ?

What do you think

SteveSandersonMS:
I think that because of the security implications of using `sudo` when installing packages, it's not something we should recommend. If you are in a situation where it fails otherwise, you might want to look into fixing your configuration so you don't need it any more. Using `sudo` when running `npm install` is literally giving root access to your machine to potentially hundreds of unknown package authors - really, don't be in that situation!

MarkPieszak:
Yeah it's definitely not ideal, I'll have to look more into how to fix all these bizarre security issues going on with this machine. It's hard to find the time! :) I'll leave this PR closed then, thanks Steve!

PR #52, opened 2016-04-26
chore(package): Add typings postinstall

MarkPieszak:


SteveSandersonMS:
We should definitely do something along these lines when we've implemented #53. Is there any benefit to doing it before then?

Looks like it presents an extra possible point of failure (will die if you don't have `typings` installed) so would be preferable not to put it in unless it's doing something essential. Right now does it do anything, given that there's no "typings" config?

MarkPieszak:
I apologize I think I posted these in the wrong order! I'll close it out and add `typings` itself first :)

Apologies Steve!

Are you going to be at Ng-Conf next week??

PR #54, opened 2016-04-28
chore: add js and jsx files to gitignore 

laskoviymishka:
This change just to prevent spam in pending changes with visual studio.
All typescript files in sample compiled to js with .map. It not really necessary to store this file in git, so better to ignore it may be

### DIFF
--- a/samples/angular/MusicStore/.gitignore
+++ b/samples/angular/MusicStore/.gitignore
@@ -4,6 +4,7 @@
 /project.lock.json
 /music-db.sqlite
 /Properties/launchSettings.json
+/typings
 
 # Obviously you don't really want your DB to go in wwwroot, but due to https://github.com/aspnet/Microsoft.Data.Sqlite/issues/188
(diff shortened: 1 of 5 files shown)


SteveSandersonMS:
Thanks for pointing this out. Would it be fair to say the root issue is that VS produces these unwanted `.js` and `.jsx` files? They are unwanted because they serve no purpose when the project is set up this way - they never get served to the browser, they just clutter up your folders and potentially confuse developers.

To avoid this, I've added `<TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>` to all the `.xproj` files, so VS will no longer produce this unwanted noise.

Does this adequately resolve the issue? It's probably preferable not to use `.gitignore` to exclude all `.js` files indiscriminately, because developers might have some reason to add their own `.js` files, and would then be confused about why they don't end up in source control.

laskoviymishka:
Good point, agree with you completely :+1: 
So it now pretty much already fixed!

PR #57, opened 2016-04-29
Fix issue: entry.main.unshift is not a function Closes #56

### LINKED ISSUE #56: Webpack default entry issue: webpackConfig.entry.main.unshift is not a function.
There is a statement inside of webpack dev middleware ts:
`webpackConfig.entry['main'].unshift('webpack-hot-middleware/client');`

Beside assuming that we have default main entry (which is not really good, but there is already TODO here), this allso assume that entry point is an array. Webpack is allow to you have this entry as a string. 
Webpack ng2 starter as example have following config:
`entry: {
    'polyfills': './src/polyfills.ts',
    'vendor': './src/vendor.ts',
    'main': './src/main.browser.ts'
  },`
In this scenario you will have error:
webpackConfig.entry.main.unshift is not a function.

laskoviymishka:
Add check for string entry - if so, replace main entry with new array of hot middleware and old entry.


SteveSandersonMS:
Excellent - thanks very much!

Issue #66, opened 2016-05-06
Prerendered SPA cookies issues


laskoviymishka:
When we have spa running under server authorization context (i.e. we have cookies authorization) we no able to run secure API calls on prerendering side.
For angular this is blocked till issue: _https://github.com/angular/universal/issues/273_ resolved
For react (react-redux template) this not working for both client and server render (isomorphic search not passing cookies), but could be fixed.

laskoviymishka:
To enabling server side cookies processing we need to path the cookie string or object from server request to JS module invoke. Currently we can do this by adding additional arg to method:
```
public static async Task<RenderToStringResult> RenderToString
```
This method further pass it throw chosen workflow (specific node services) and prepare context object which be accessible in JS rendered module.
This is quite simple feature (from server side perspective), so i could create it.
For the client side modification (template for ng2 still blocked, react template need some deep refactoring), better to create separate issues.

SteveSandersonMS:
If I understand correctly, the basic thing missing from NodeServices here is the ability to pass through additional arbitrary information from the `asp-prerender-*` tag helpers into the JS boot function. If that facility existed, you could pass through whatever authentication information was relevant to your application, which might be:

 * The authenticated username, if you're using ASP.NET auth
 * A JWT token or similar, if you're using some external auth mechanism
 * Some information from the user's `Session` store, if you have some custom auth system or need to pass through something like "what features this user has paid for" etc.

Do you agree?

Closing this in favour of the more general #107

laskoviymishka:
@SteveSandersonMS not really, it mean that we have no ability to operate with request cookies in prerender process. If you have regular MVC cookies auth, and you try to fetch secure resource it will work on client, but fail on server render. This issues simillar but have really different background. This behaivior should populated by default.

laskoviymishka:
JWT token is impossible in server render process (it obvious couse we don't have no connect to any client storage at server).
Sessing and cookies could be passed as `asp-prerender-model` but is it really best option?

SteveSandersonMS:
I understand you are looking for specific cookie support, rather than relying on the developer to pass through the information manually. Filed #130 to track that.

jeinz11:
I need this feature badly because im using mvc cookie authentication....
please help

jeinz11:
Cookie support please

jeinz11:
@laskoviymishka  hi can i use this right now?

     public static async Task<RenderToStringResult> RenderToString

and pass args with cookies please help

PR #67, opened 2016-05-06
wip: Cookies issue and payload provider

laskoviymishka:
Work in progress.
Created basic sample for reproducing issue #66 also some refactor to #68.


laskoviymishka:
Implement basic cookie pushing from server to boot function. Also refactor a little bit prerender process to enable payload passing from server request to boot function also (related to issue #68)
CC @SteveSandersonMS

angelyordanov:
Hey @laskoviymishka great work!
Just a few notes though.
1) Changing the code style of files you are not editing makes it harder to track your changes.
2) Isn't it better the new method in the `Prerenderer` to accept the `PrerendererOptions` as an argument, not just forward the call to the `INodeServices`. After all its idea is to be a wrapper of the `aspnet-prerendering` module.

laskoviymishka:
Hi @angelyordanov, thx for feedback
1) Visual studio change code style, and I didn't mention it, my bad. IMHO this repo code should be adjusted to common ms code style rules for c#. Right now code unreadable.
2) I don't see any reason why this static `Prerenderer` class should be here, i should remove it at all. Right now it not used at all. May be it better to extract prerendering functionality from `PrerenderTagHelper` to some `IPrerenderer` which will be registred in `ServicesCollection` so it could be accessed anywhere, not just tag helper. Any thoughts on this?

This is still in progress, so any feedback is highly appreciated.

angelyordanov:
I tried to imagine what would an `IPrerenderer` interface look like, and what would be the benefit of replacing it through the DI subsystem and the only thing I could think of is if one wants to substitute the `aspnet-prerendering` module with its own `renderToString ` implementation(removing domain-tasks, webpack building, etc..). But then we need to somehow generalize the `PrerenderTagHelper`'s attributes and the way they are passed to the `IPrerenderer`.

Anyway, if one wants to do that he can just copy the `PrerenderTagHelper` + `Prerenderer` and start from there. Its not that much code.

So, You are right, the `Prerenderer` is an implementation detail and is probably best removed or made internal.

laskoviymishka:
Refactor prerenderer a little bit:
- Introduce prerenderer interface
- Extract classes to separate files
- Remove static accessor to prerenderer
- Clean prerender options
- Remove magic callback setter to prior manual population of args
- Add basic payload population for ng2 auth sample

So now basic setup is not require add node services anymore, but you could modify some prerendering options with following code in your startup:

```
public void ConfigureServices(IServiceCollection services)
{
          services.AddPrerender(
                    r => new PrerenderOptions
                    {
                              /* your project specific options goes here */
                    });
}
```

Also not it much easier to pass some data from server direct to your boot JS function. Here is sample:
```
public void ConfigureServices(IServiceCollection services)
{
          services.AddPrerender(
                    r => new PrerenderOptions
                    {
                              PayloadAccessor = context => new object[]
                              {
                                  "hello i'm payload from server",
                                  "my is also from server, but i'm, unicorn",
                                  new { name = "Pink", type = "Fluffy", race = "Unicorn" }
                              }
                    });
}
```
In this callback scope there avaible service provider and http context object, so it could be pretty easy to extract this population to somekind of service. Here is sample:

```
public void ConfigureServices(IServiceCollection services)
{
          services.AddPrerender(
                    provider => new PrerenderOptions
                    {
                              PayloadAccessor = provider.GetServices<IInitialStateService>().PopulatePayload
                    });
}
```

Folks, any thoughts on this?

laskoviymishka:
Merge rc2 upgrade into this quite painfull, better to recreate it as new PR.

jeinz11:
how to use this addPrerender it says the addPrerender does not exist thanks

laskoviymishka:
This pr was not merged and proposed api is not in mainline yet

jeinz11:
oh no...im stuck right now passing cookie to my api controller during SSR do you have any solutions? thanks

laskoviymishka:
There is a solution but with a slightly different api. It should be described in docs page

jeinz11:
thanks i was trying to pass it in my redux state by  using this code

        const store = configureStore({ currentUser: { isUserLoggedIn: false,userCookie:"asdasdasdsdas" } });

but when the time i retrieve... it doesnt change the value  when doing SSR  please help huhuhuhu

jeinz11:
@laskoviymishka  please help me sir

laskoviymishka:
You should use `asp-prerender-data` attribute of tag helper to pass cookie data to your js code. Then in js code you should register some opaque token for that and that it.

laskoviymishka:
On client side there is 
```
export interface BootFuncParams {
    location: url.Url;          // e.g., Location object containing information '/some/path'
    origin: string;             // e.g., 'https://example.com:1234'
    url: string;                // e.g., '/some/path'
    absoluteUrl: string;        // e.g., 'https://example.com:1234/some/path'
    domainTasks: Promise<any>;
    data: any;                  // any custom object passed through from .NET
}

```

Object available. This object may contain your HttpContext.Request.Cookies

PR #69, opened 2016-05-10
feat(ng2-template): upgrade to rc1

### LINKED ISSUE #62: Angular2 rc1 upgrade
Angular team have published first release candidate. We should upgrade template for it (and may be music store sample, but this part could be real pain).

laskoviymishka:
Closes #62
Only for template. Music store still on beta.17


SteveSandersonMS:
Thanks very much for this! I've merged commits 0571e4 and 57091f (rebased onto `master`, which now targets ASP.NET Core RC2). The other commits are part of other pull requests so they can be handled separately.

Issue #72, opened 2016-05-16
process.env.ASPNET_ENV undefinied


laskoviymishka:
When running project with web pack dev middlewere process.env.ASPNET_ENV always undefinied.

SteveSandersonMS:
Closing in favour of #73

PR #73, opened 2016-05-16
Fix issue with ASPNET_ENV always undefinied

laskoviymishka:
Follow aspnet env from `IHostingEnviroment` interface to node process in prerender and webpack flows.


laskoviymishka:
@SteveSandersonMS merge rc2 migration, please review

SteveSandersonMS:
I'd be happy to merge this if it's needed, but investigating just now, I wasn't able to repro the original issue (#72). Maybe it's been fixed in .NET Core RC2, or maybe I'm misunderstanding the requrement.

Is the issue that you want the `ASPNETCORE_ENVIRONMENT` environment variable to be made available in your Node.js code (i.e., as `process.env.ASPNETCORE_ENVIRONMENT`)? If so, as far as I can tell, that already happens by default, because the Node.js child process gets all the environment variables from the parent process.

I just checked this both on OS X and Windows (and on Windows, both when launching with `dotnet run` and inside VS/IISExpress). In all cases, I find that all the environment variables, including `ASPNETCORE_ENVIRONMENT`, are showing up in the Node.js child process.

Can you let me know if I'm misunderstanding the requirement? If not, maybe it was a bug in .NET Core RC1 that was fixed in RC2.

SteveSandersonMS:
@laskoviymishka, sorry to hassle you, but do you have any thoughts on my questions above? I'll keep things tidy by closing this if there are no further requirements.

laskoviymishka:
Sorry I away from my PC :(
You can close it for now. If I find anything new I just reopen this PR or create new one

SteveSandersonMS:
OK, thanks for confirming!

Issue #89, opened 2016-05-23
Full framework support


rmja:
Hi, will there be `net451` support?

SteveSandersonMS:
Merged your pull request - thanks!

PR #90, opened 2016-05-23
net451 support and netcoreapp1.0->netstandard1.5 for libraries

rmja:
Resolves https://github.com/aspnet/JavaScriptServices/issues/89

### DIFF
--- a/src/Microsoft.AspNetCore.AngularServices/project.json
+++ b/src/Microsoft.AspNetCore.AngularServices/project.json
@@ -13,7 +13,9 @@
     "Microsoft.AspNetCore.SpaServices": "1.0.0-*"
   },
   "frameworks": {
-    "netcoreapp1.0": {
+    "net451": {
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
Thanks very much! This is now merged.

If you discover anything not working on .NET 4.5.1, please let me know. My testing has been all about .NET Core.

zigzag95:
Thank you
Nuget ready?

laskoviymishka:
What about netsyandart 1.6? It's escalate quickly :)

xabikos:
Actually instead of targeting netstandard1.6 we should try and target a lowest number if possible (1.3) as this would give much more compatibility especially now that full .net framework is supported. Of course I don't know if it's possible based on api's the library uses.

laskoviymishka:
Well it could, we need investigate possibilities of such targeting's

PR #94, opened 2016-05-24
bug: Angular 2 not working with InputOutputStream

### LINKED ISSUE #93: Angular 2 in debug mode not working with InputOutputStream
### Steps to reproduce

1. Open solution
2. Select Angular2Spa from templates as startup project
3. Change `HostingModel` by adding following code in `ConfigureServices` method:
```
            services.AddNodeServices(new NodeServicesOptions
            {
                HostingModel = NodeHostingModel.InputOutputStream
            });
```
4. Open any page
5. See exception
```
JsonReaderException: Unexpected character encountered while parsing value: A. Path '', line 0, position 0.
```

### The issue

Page not load, and exception accured

```
JsonReaderException: Unexpected character encountered while parsing value: A. Path '', line 0, position 0.
```

### Further technical details

This is pretty easy issue - right now `InputOutputStreamNodeInstance` wait for first recieved data from Node, and claim it as boot function response. With ng2 it not so, since in development mode it everytime produce message: `Angular 2 is running in the development mode. Call enableProdMode() to enable the production mode.`
To fix this issue we need implement some kind of filtering our result. Root couse placed inside `OnOutputDataReceived` method in `InputOutputStreamNodeInstance` class.

laskoviymishka:
Closes #93
We need to ensure that output recieved data is real boot function response, and not just regular console.log from node process.
This is pretty dumb, but it will fix basic behaivior which described in #93.
If anyone have some ideas how it could be improved, any feedback would be GREAT!

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/HostingModels/InputOutputStreamNodeInstance.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/HostingModels/InputOutputStreamNodeInstance.cs
@@ -48,7 +48,7 @@ namespace Microsoft.AspNetCore.NodeServices {
         }
 
         protected override void OnOutputDataReceived(string outputData) {
-            if (this._currentInvocationResult != null) {
+            if (this._currentInvocationResult != null  && outputData.Contains("{")) {
                 this._currentInvocationResult.SetResult(outputData);
             } else {
                 base.OnOutputDataReceived(outputData);


laskoviymishka:
update to latest master. Pinh @SteveSandersonMS

SteveSandersonMS:
Thanks for submitting this. Looking for a `{` character may resolve the issue in the specific case of #93, but as you point out, leaves open the underlying issue which is that the `InputOutputStream` transport will trip up if it sees unexpected data on the stream (and that unexpected data might include a `{`).

The `InputOutputStream` transport was never meant for real use - it was just to show that other types of transport were possible.

I've just pushed the new `Socket` transport, which is a robust and more generalised stream-based transport. `Socket` will soon become the new default (replacing `Http`) once I've got a bit more feedback about it, because it's faster and has more capabilities than `Http`.

Now `Socket` is in place, I'm going to remove `InputOutputStream` altogether, as it no longer serves any purpose. If anyone has any objections please let me know.

If you don't mind, then, I'll close this as it will no longer be applicable.

PR #100, opened 2016-05-27
In order to use this package in traditional .NET setups and NuGet env…

pauldotknopf:
…ironments, there needs to be specific environments created for each version of .NET.

I'm not seeing another way to do this. I may be wrong.

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/Configuration.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/Configuration.cs
@@ -1,3 +1,4 @@
+#if DI_CONFIGURATION
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.PlatformAbstractions;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
To help me understand better, could you describe what about the existing setup is *not* working correctly on "traditional .NET setups and NuGet environments"? Can you describe it like a bug repro, as in what you expect to work that isn't? I'm pretty much entirely focused on .NET Core environments so I might not have tried whatever thing you did.

> there needs to be specific environments created for each version of .NET. I'm not seeing another way to do this. I may be wrong.

My expectation was that there was no need to declare `#if` conditions for every combination of .NET environment, because these are resolved at compile-time. Just declaring support for .NET 4.5.1 and having relevant `#if` blocks for `NET451` means that the build output includes a .NET 4.5.1-compatible binary, which will be compatible with .NET 4.6.x (etc.) by virtue of .NET's own back-compatibility.

But please let me know what's not working in your environment and we can figure out for definite what's needed for correct support.

SteveSandersonMS:
I'll close this as the discussion seems to be inactive, but if anyone feels there is a problem to be solved here and can specify details, please reopen or post another issue!

PR #108, opened 2016-06-01
feat(prerender): ability to push payload and model to boot func

### LINKED ISSUE #66: Prerendered SPA cookies issues

### LINKED ISSUE #68: Add prerender pipeline hooks for better extensibility
This need some design, so i address it to @SteveSandersonMS, if you good with such approach i could implement this.
So for better extensability it will be realy nice to have some pipeline injects to manage some additional variables passing from asp.net server directly to server JS (like some auth info beside cookie, or may be initial state of Redux store, so we don't have to grab it from web api via additional http requests).
For example there is could be some `PrerenderOptions` class which could be extended via calling `AddPrerender` method in `ConfigureServices` method of application startup class. This method just register some hepler interface to run prerender manually and setup prerender opions.
This options could be something like
```
 public class PrerenderOptions
 {
        public Action<Request, NodeInvocationInfo> BeforeRenderModule{ get; set; }

        public NodeServicesOptions() {
            this.HostingModel = _ => { };
        }
 }
```

Inside on `NodeInvocationInfo` options already `Args[]` property which passing directly to the JS Module invocation, so we could push anything to our JS code from server via this hook.

This is kinda realted to issue #66 (both require prerry much same refactoring, to establish clearer pipeline for prerendering process).

### LINKED ISSUE #107: Proposal: Add ability to pass @Model to prerendered boot func
It would be nice to see some ability to pass extension model for our prerendered components. It could be very helpfull for some redux/flux apps (we could populate initial state with them). And it could reduce total overhead of loading some info via Http requests.
Proposal api looks pretty simple. In tag helper add populate new attribute:
```
<app asp-prerender-module="ClientApp/boot-server"
     asp-prerender-model="Model"
     asp-prerender-webpack-config="webpack.config.js">Loading...</app>
```

Inside boot func there is new optional parameter, which popuated from `asp-prerender-model` attribute:
```
export interface BootFunc {
  (params: BootFuncParams, model: any): Promise<RenderToStringResult>;
}
```
Such feature would be really good for building some dynamic parts for exist applications (not whole application handled by some JS-framework).

laskoviymishka:
- closes #68 by adding payload provider
- closes #66 by pushing context cookies to boot func
- closes #107 by adding asp-prerender-model attribute

Implemented API:

**Payload provider**
Right now there is an ability to implement some common `PayloadProvider` which could add some neccesary information to every prerender func call. To implement such provider you could use following API:
```
services.AddPrerender(provider => new PrerenderOptions
            {
                PayloadProvider = context => new[] { new { Food = "Borsch", Drink = "Vodka" } },
            });
```
`provider` is just regular `IServiceProvider` with all registred services.
`context` is just a regular `HttpContext` with all required info for your request.
After specifying these `PayloadProvider` you could access this array of object via params object:
```
import * as aspnet from 'aspnet-prerendering';
let boot: aspnet.BootFunc = function (params, model) {
  console.log(params.payload);
};
```

 **Accessing model**
To access model you could add following code to your `View.cshtml`
```
<app asp-prerender-module="ClientApp/boot-server"
     asp-prerender-model="Model"
     asp-prerender-webpack-config="webpack.config.js">Loading...</app>
```
And then access to passed model:
```
import * as aspnet from 'aspnet-prerendering';
let boot: aspnet.BootFunc = function (params, model) {
  console.log(model);
};
```

 **Cookies**
Now it possible to catch aspnet request cookies at server prerender. You may perform these with following code:
```
import * as aspnet from 'aspnet-prerendering';
let boot: aspnet.BootFunc = function (params, model) {
   console.log(params.cookies);
};
```

cc @SteveSandersonMS 

Any feedback would be great for these one.

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/INodeInstance.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/INodeInstance.cs
@@ -6,7 +6,7 @@ namespace Microsoft.AspNetCore.NodeServices
     public interface INodeServices : IDisposable
     {
         Task<T> Invoke<T>(string moduleName, params object[] args);
-
+        Task<T> Invoke<T>(NodeInvocationInfo nodeInvocationInfo);
         Task<T> InvokeExport<T>(string moduleName, string exportedFunctionName, params object[] args);
     }
 }
--- a/src/Microsoft.AspNetCore.SpaServices/Configuration.cs
+++ b/src/Microsoft.AspNetCore.SpaServices/Configuration.cs
@@ -1,0 +1,17 @@
+﻿using Microsoft.AspNetCore.SpaServices.Prerendering;
+using Microsoft.Extensions.DependencyInjection;
+using System;
+
+namespace Microsoft.AspNetCore.SpaServices
+{
+    public static class Configuration
+    {
+        public static void AddPrerender(this IServiceCollection serviceCollection) => AddPrerender(serviceCollection, _ => new PrerenderOptions());
+
+        public static void AddPrerender(this IServiceCollection services, Func<IServiceProvider, PrerenderOptions> optionsFactory)
+        {
+            services.AddTransient(typeof(PrerenderOptions), optionsFactory);
+            services.AddSingleton(typeof(IPrerenderer), serviceProvider => new Prerenderer(serviceProvider));
+        }
+    }
+}
--- a/src/Microsoft.AspNetCore.SpaServices/PrerenderOptions.cs
+++ b/src/Microsoft.AspNetCore.SpaServices/PrerenderOptions.cs
@@ -1,0 +1,12 @@
+﻿using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.SpaServices.Prerendering
+{
+    public class PrerenderOptions
+    {
+        public Func<HttpContext, object[]> PayloadProvider { get; set; } = context => new object[0];
+        public string FileName { get; set; }
+        public string ExportedFunctionName { get; set; } = "renderToString";
+    }
+}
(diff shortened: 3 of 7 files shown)


laskoviymishka:
update with latest master

SteveSandersonMS:
Thanks again for raising this. The need to pass additional data into the JS boot function is an important requirement.

As per our conversation on #107, I think this requirement is now covered by the new `asp-prerender-data` attribute. This low-level API can be used to build any other data-passing API on top of, however a developer wants, although I suspect that `asp-prerender-data` on its own would be enough in typical cases.

I'll close this as I think the basic need is covered. However I do appreciate that your pull request here would introduce additional features that for some developers would be useful (at the cost of API expansion). I'll be looking out for how developers use `asp-prerender-data` in practice and what patterns become commonplace around it. In the future, if we hear that people keep wanting this sort of payload provider pattern and don't feel comfortable implementing their own, we can certainly consider having a native one. Hope that makes sense, and thanks again for helping push this design forwards!

laskoviymishka:
I understand your concern regards payload provider, but what about cookies? This is must have feature. Without this one any serious application with server prerender become useless.

SteveSandersonMS:
I wrote some thoughts about cookies here: https://github.com/aspnet/JavaScriptServices/issues/130

jeinz11:
Developer with cookie authentication needs this... right now my first request all cookies are null and it cant pass my authorization thanks

PR #111, opened 2016-06-02
upgrade netstandard, point to 1.6 (according to latest MVC)

laskoviymishka:
- add missed projects to solution
- fix broken build for them
- reorgonize solution folders for sample (adjust to physical paths)

Not really sure regards to netstandard version, but i was unable to build and run projects with current version (1.3). After a little research i found that latest mvc pointed to 1.6 already (see _https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc/project.json#L40_), so probable we should point these version also (we have dependency on MVC package in SPAServices). NodeServices could still be pointed to 1.3.

Also add missed sample to solutions, and adjust correct project.json to them. Reorgonize solution structure for sample, so now it same as physicall structure


SteveSandersonMS:
> but i was unable to build and run projects with current version (1.3)

Can you specify what the problem was? It's working OK for me on `netstandard1.3` as-is. Obviously it's preferable only to depend on the minimum netstandard version needed.

laskoviymishka:
@SteveSandersonMS In Visual studio it not resolve all dependencies from `Mvc.*`. (it able to run throw console)

SteveSandersonMS:
This is now partially merged - the updated `.sln` file, new `.xproj`, etc., are in `master` as 7e51fc44801cfdb7d01c26e651f79c5ec0bbf261.

As for the changes to netstandard version, I haven't merged those because I still haven't been able to establish that they are necessary. I'm finding the current code in `master` already opens and builds correctly in VS (specifically, VS 2015 Update 2, with the ASP.NET Core 1.0 RC2 tooling installed).

> And current minimum version of netstandart for MVC is 1.6 as you can see in PR description link

Yep, I see that. And for this reason there's no major drawback to taking a dependency on netstandard 1.6 from the projects that depend on MVC. But I would still want to understand the reason for a change before making it. You mentioned that your version of VS had trouble resolving some dependencies from MVC, but I've not been able to repro that. We should at least understand what's going on and why there's an inconsistency there. Is it possible that when you changed netstandard version on your machine, you just (in effect) forced it to reset its existing downloaded dependency graph and hence resolved some problem that was just temporary on your machine anyway? Can you repro the netstandard1.3 problem on a clean repo? Let's be sure we understand the underlying cause before taking action. Hope that seems reasonable!

SteveSandersonMS:
@laskoviymishka, sorry to hassle you, but do you have any thoughts on my questions above? I'll keep things tidy by closing this if there are no further requirements.

laskoviymishka:
Same on this. I was include nightly builds of mvc, which supports only 1.6. But before goes with preview1 this should be applied.

Issue #117, opened 2016-06-07
Provide a React template in ES6 syntax


xabikos:
I think it's quite essential to provide a React Redux template in ES6 syntax. In general TypeScript is getting more and more popular among Angular developers. But in the React community ES6 is more popular.

SteveSandersonMS:
Closing as duplicate of #118

PR #118, opened 2016-06-07
Add React Redux template in ES6 syntax #117

xabikos:
It's an exact copy of the template React Redux but in ES6 syntax instead of TypeScript.

If this is merged then we should include it in the corresponding yeoman generator.


SteveSandersonMS:
Thanks for submitting this.

It looks good, but I'm not sure about whether it should be merged into the core repo. There are already a lot of combinations of scenarios (and we will be adding Aurelia soon too) - having TypeScript as a common language at least provides one point of consistency. If we did a pure-ES2015 version of the ReactReduxSpa template, then we'd also need to write and maintain pure-ES2015 versions of all the others too. And then somebody would want Flow and CoffeeScript versions too :)

Would you be happy with publishing your ES2015 ReactReduxSpa sample/generator separately?

SteveSandersonMS:
To be clear, I do think what you've done is great and valuable, and definitely should be published - people will want to use it. I'm just trying to limit this repo to avoid it exploding with matrices of combinations of frameworks and languages :)

laskoviymishka:
@SteveSandersonMS but why not? Made flexible yo generator is really helpfull for comunity. Especially now, when a lot of fresh dev start to look on .net core. It little bit hard to find any good template for modern SPA app with aspnet now. This is most up-to date place. So may be it really good to provide as expressive as possible generator? This is really nice feature. 

Current templates structure is not really looks well from extension point, so any new template will require a lot of duplicates. But may be it worse to research how it could be improved?

SteveSandersonMS:
> but why not?

The maintenance cost. We already cover a lot of frameworks - extending that to cover all combinations of frameworks and multiple code languages would be impractical, and supporting a random subset of combinations would make no sense. There's only a certain amount of time to work on this whole project, and its primary goal isn't to be an exhaustive repository of starter kits.

I totally agree that @xabikos's template is valuable, and it should be published. But it can be published as a separate generator - that's the whole idea of Yeoman generators being in independent packages.

**Suggestion:** To ease discoverability, we could start a wiki page in this repo giving links to related Yeoman generators. Then whatever combination of related generators are published by the community can be listed there, without creating the expectation that all combinations will be implemented and maintained here.

@xabikos - does that sound like a good approach to you?

laskoviymishka:
@SteveSandersonMS idea with wiki of generators is really great :+1: 
Maintenance cost could be huge for sure. Especially with current structure when each project tempalte made as separate application. 
But having everything that you need in one place is huge benefit, and comunity may handle some of maintenance effort :smile:. You are not alone!

xabikos:
@SteveSandersonMS I understand your point and I agree about the maintenance cost. Personally I did spend the time of developing this template as I am going to give a [presentation](https://skillsmatter.com/conferences/7235-progressive-dot-net-tutorials-2016#program) in London in a couple of weeks about isomorphic applications in .NET and TypeScript gives me a hard time.

Based on that I think it's a pity to concentrate only in TypeScript as this will give a negative feeling to developers outside Microsoft ecosystem to jump into the platform. For Angular 2 of course it makes sense but for React I don't think so.

Now about the wiki it's a good idea but I would suggest something different. In order to keep this repo small maybe the templates and samples should be moved in a different repo where more people will be able to contribute and provide templates. So this repo will contain just the core projects of the library.

In all cases I will probably create a repo in my personal account and add this template.

simonwoodhead:
I agree that too many generators could be confusing but I am not sure it is a good idea to use TypeScript as a point of consistency across the different templates. Maybe each template should be as close as possible to the defaults used by that frontend language? (So in the case of React, ES2015, rather than TypeScript.)

xabikos:
I don't want to argue the decision of just using TypeScript but I totally respect it. I don't want to go further with the discussion in this pull request. This week I will add this template in my personal github account and then close this.

SteveSandersonMS:
Thanks for the additional comments. As per @xabikos's comment, I'll close this - he can publish his template elsewhere, and we can reference it.

As for the general question of TypeScript vs ES2015, we can remain open to any possibility in the future. Right now the scope of this repo needs to be constrained so quality can be kept high. Note that these are not *general* SPA starter kits - they are specifically for ASP.NET Core apps, which means they are for C# developers specifically. Typically we hear so much positivity about TypeScript from C# devs that it would seem very strange not to use it by default. If we continue to get feedback over time that ASP.NET Core devs who are using React on the client prefer pure ES2015, we'll definitely allow for that.

xabikos:
[Here](https://github.com/xabikos/aspnetcore-react-redux-template) you can find the ES2015 template of React Redux SPA application

emzero:
TypeScript is cool but I do prefer ES2015 too because of 2 reasons.

1) It's the standard.
2) Nowadays everything is changing too fast. Everyday new client-side libraries are emerging. In order to use TypeScript you need the typings which are not usually included in all libraries and makes everything much more annoying.

Just my 2 cents.

thangchung:
@xabikos I was looking for the ES6 version in awhile. I like your works.  👍

PR #161, opened 2016-07-05
Makes it possible to use absolute paths for publicPath in weback.config

tobiasr:
- Parses publicPath as an URL and returns the path of that URL to be able to cope with absolute URLs


SteveSandersonMS:
Thanks - looks great :)

PR #177, opened 2016-07-13
Added OnBeforeStartExternalProcess callback 

### LINKED ISSUE #20: Don't force Node to be on the PATH
From what I can read, the code currently assumes that Node is on the PATH. I think it's a good idea not to bundle a specific version of Node with this package, but it would be great if it was possible to specify where the Node executable is. Would make it easier to deploy multiple sites using this to the same server where the different sites use different Node versions.

thunder7553:
Added OnBeforeStartExternalProcess callback which to NodeServicesOptions (and OutOfProcessNodeInstance, SocketNodeInstance and HttpNodeInstance) to configure environment of the node.exe process to be started, and the path to the node executable itself. Fixes #20

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/Configuration/NodeServicesOptions.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/Configuration/NodeServicesOptions.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.NodeServices
             HostingModel = DefaultNodeHostingModel;
             WatchFileExtensions = (string[])DefaultWatchFileExtensions.Clone();
         }
-
+        public Action<System.Diagnostics.ProcessStartInfo> OnBeforeStartExternalProcess { get; set; }
         public NodeHostingModel HostingModel { get; set; }
         public Func<INodeInstance> NodeInstanceFactory { get; set; }
         public string ProjectPath { get; set; }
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
Thanks for this! I've now merged it. I hope you don't mind, but I've also changed the API.

I was not keen to expand the set of constructor parameters for all the hosting models (even if it's a new optional parameter, it's still something that everyone calling the constructor has to see and figure out the relevance of).

But since it was a constructor parameter, the only way to make use of it would have been when calling `new` directly (and not just when receiving an `INodeInstance` through DI). So if a developer is already doing that, it's hardly any extra work to define a subclass of whatever hosting model they are instantiating. So, this can conveniently be done as a virtual method instead of a constructor parameter.

You can now do this:

    public class MyHttpNodeInstance : HttpNodeInstance
    {
        protected override ProcessStartInfo PrepareNodeProcessStartInfo(string entryPointFilename, string projectPath, string commandLineArguments)
        {
            var startInfo = base.PrepareNodeProcessStartInfo(entryPointFilename, projectPath, commandLineArguments);
            // Now modify 'startInfo' however you want to here
            return startInfo;
        }
    }

I know this is slightly more code than would have been needed with the constructor param, but it has the benefit of not interjecting an extra concept to think about for developers who are just trying to instantiate a node instance without wanting to override the process start info in any way. Hope that's OK - let me know if you disagree!

Issue #53, opened 2016-04-26
Remove tsd - update to typings


MarkPieszak:
To do: Update sample projects to use `typings` not deprecated `tsd`.
Also update TS to 1.8+

emzero:
+1

MarkPieszak:
I'll start working on this one this week, sorry for the delay

SteveSandersonMS:
Do you think this is still worth doing? The TypeScript people have announced that, as of v2.0, TypeScript will have native support for acquiring .d.ts files directly from NPM, under the `@types` namespace that they will be managing. This will supersede the use of tsd/typings.

I don't know how soon that's going to ship, but unless the use of `tsd` is really problematic for anyone in the short-term, we might decide it's a better use of our resources to leave things as-is until TypeScript v2 ships.

SteveSandersonMS:
Oh, for more information about this, see https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

MarkPieszak:
I was actually about to comment on this last night about that! You read my mind :)
It's up to you! Switching from `tsd` to `typings` is real easy as well.

Very excited for @types... being able to have all those dependencies listened in the package.json will make so much simpler.

Should I close this one out or rename the title?

SteveSandersonMS:
I'll close this. We can pick up the @types stuff when TypeScript 2 ships - there's nothing for us to do until then. Thanks for your input!

PR #179, opened 2016-07-13
Upgraded to typings from tsd. Moved URL workaround and load it with typings

anwalkers:
Upgraded to typings from tsd. You will need to install typings globally

`npm install -g typings`


SteveSandersonMS:
Thanks @anwalkers for submitting this! It's definitely great to keep the templates on an up-to-date technology stack.

A while ago in #53, @MarkPieszak and I were planning to update to Typings, but decided to put in on hold when the TypeScript team announced that with TypeScript v2.0, they would have native support for acquiring `.d.ts` files from NPM, so it will no longer be necessary to use either TSD or Typings for this in the templates.

I don't know how soon TypeScript 2.0 is going to ship, but I'm guessing it must be pretty much imminent - the beta has been out for a long time, and the roadmap planning has already moved on to 2.1.

Given the likelihood of a TS 2.0 release soon, and that we'll update to use its native support for acquiring `.d.ts` files at that time, would you agree it might be better to minimise the amount of churn by sticking with `tsd` until then? I don't think it will be hugely beneficial to encourage a change to this for what might be just a few days or weeks - any time spent tracking down new issues or incompatibilities would be obsolete very soon, and could better be spent enhancing other features.

What do you think? Would it be OK to wait until TS 2.0 ships? Obviously for your own project you've already put in Typings support so it isn't going to affect you, but I suspect that for others, having a simpler story about moving to native TS 2.0 support when available could be a preferable arrangement.

MarkPieszak:
@SteveSandersonMS So I just updated the Universal project to use the new @types, but it's up to you whether you think we should wait for the official release or use typescript@beta and start using it already. It's definitely really nice to just have the package.json contain everything now, love 2.0 so far! Nice job by the TS team :)

anwalkers:
Sounds good.

PR #184, opened 2016-07-15
Adding support for capturing the output of a node instance for custom…

pauldotknopf:
… logging implementations.

Not all implementations rely on stdout for logging and error messages.


laskoviymishka:
Great point, best way to handle it IMHO to replace every Console.WriteLine to ILogger.Write, and inject logger from DI. Does it make sense?

pauldotknopf:
The ```INodeInstance```s are created via ```new``` (see [here](https://github.com/aspnet/JavaScriptServices/blob/eec370e9389abc6ef32fc3b5109f398f75d023e6/src/Microsoft.AspNetCore.NodeServices/Configuration/Configuration.cs#L49)). No room for DI there.

pauldotknopf:
Are you referring to using the Microsoft libraries for logging?

laskoviymishka:
Yep, it is. And yep it need to be refactored :)

Issue #172, opened 2016-07-12
Angular2's (click) events doesn't work on IE 9


skamlet:
The `(click)` event doesn't seem to work with a project generated with 'Aspnetcore Spa' generator.
 => can be tested on the "Counter" page

Though creating a "normal" node project (following the instructions from the Angular2's quick start) doesn't cause this problem.

Tested on IE 9 on Windows 7

laskoviymishka:
This should be addressed to angular2 team. I think there is only IE 10 support from box. Everything else require polyphils or god helps. For IE 9 i assume that only god could help you.

skamlet:
Thing is, with a freshly created angular 2 project running on node, I don't have this problem... But yeah, for IE 9 i'll also assume god is the only help :P...

laskoviymishka:
What does console say? Are you add polyphill? Can you provide more details about what difference between those 2 projects? Webpack config etc.
And also pray to any god you know. It also may help you :ship:

MarkPieszak:
Angular2 should work on IE9 (unless the latest rc broke in it). 

Do you have es6-shim or better yet, core-js shim?

skamlet:
@MarkPieszak I tried with both, no success.

@laskoviymishka:
The **NOT working** project is using:
```
webpack.config.vendor.js:
var path = require('path');
var webpack = require('webpack');
var ExtractTextPlugin = require('extract-text-webpack-plugin');
var extractCSS = new ExtractTextPlugin('vendor.css');
var isDevelopment = process.env.ASPNETCORE_ENVIRONMENT === 'Development';

module.exports = {
    resolve: {
        extensions: [ '', '.js' ]
    },
    module: {
        loaders: [
            { test: /\.(png|woff|woff2|eot|ttf|svg)$/, loader: 'url-loader?limit=100000' },
            { test: /\.css/, loader: extractCSS.extract(['css']) }
        ]
    },
    entry: {
        vendor: [
            'bootstrap',
            'bootstrap/dist/css/bootstrap.css',
            'style-loader',
            'jquery',
            '@angular/common',
            '@angular/compiler',
            '@angular/core',
            '@angular/http',
            '@angular/platform-browser',
            '@angular/platform-browser-dynamic',
            '@angular/router',
            '@angular/platform-server',
        ]
    },
    output: {
        path: path.join(__dirname, 'wwwroot', 'dist'),
        filename: '[name].js',
        library: '[name]_[hash]',
    },
    plugins: [
[...]

PR #211, opened 2016-07-22
fix angular's (click) events not being triggered on IE9

skamlet:
Fix for angular's (click) event not being triggered on IE9.
Linked with following issue: #172


SteveSandersonMS:
Thanks - merged :)

PR #240, opened 2016-08-03
chore(github): Add github templates

MarkPieszak:
Help clean up issues & pull-requests.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,46 +1,25 @@
-**Note: for support questions, please use one of these channels:** https://github.com/aspnet/Home/blob/dev/CONTRIBUTING.md#filing-issues. This repository's issues are reserved for feature requests and bug reports.
-
 * **I'm submitting a ...**
 - [ ] bug report
 - [ ] feature request
-- [ ] support request => Please do not submit support request here, see note at the top of this template.
 
-* **What Modules or Template are related to this Issue?**
+* **What Module(s) or Template are related to this Issue?**
 - [ ] AngularServices
 - [ ] NodeServices
 - [ ] ReactServices
 - [ ] SpaServices
+
 - [ ] Angular2 Spa
 - [ ] Knockout Spa 
 - [ ] ReactRedux Spa
 - [ ] React Spa
 
 
-* **Do you want to request a *feature* or report a *bug*?**
-
-
-
 * **What is the current behavior?**
-
-
-
-* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** by creating a github repo.
-
-
+* 
 
 * **What is the expected behavior?**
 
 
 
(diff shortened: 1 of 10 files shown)


SteveSandersonMS:
Thanks for the suggestion. I'd be happy to have some kind of `PULL_REQUEST_TEMPLATE.md` at some point, but would have to think carefully about what should go into it. I'm not a fan of the bureaucratic style used here. Having been on the other side of one of this style of forms recently, I'd say it made me (as an issue submitter) feel a little patronised. I'd like to preserve a more welcoming and informal community feel for as long as possible :)

So is it OK if we don't add this in the short term?

BTW thanks nonetheless for the suggestion and for your many contributions!

MarkPieszak:
Of course, .NET guy for almost 10 years now! Just want to help get Core working with everything, thankful you started all this! I tried a while back, but in alpha it was just too hard to find any documentation !

And you know what, you're right. Want me to try and really narrow it down?

Maybe for the issue template we can make it much much cleaner like so:

----
 
* **Is this JavaScriptServices issue involving a Module(s)?**
- [ ] AngularServices
- [ ] NodeServices
- [ ] ReactServices
- [ ] SpaServices

* **Is this issue involving one of the Templates?** *
- [ ] Angular2 Spa
- [ ] Knockout Spa 
- [ ] ReactRedux Spa
- [ ] React Spa

* **Please describe the issue in as much detail as you can:**


----

SteveSandersonMS:
Thanks again for trying to think this through.

For now at least, my view is that the benefits of being approachable to newcomers (engaging with their questions/suggestions in whatever format they see best to express them) outweighs any need to streamline the project's administrative processes.

So are you OK if I don't put a `PULL_REQUEST_TEMPLATE.md` on this repo (at least not one that looks like a form to be filled out) for now?

MarkPieszak:
Yeah that's fine you can remove it!
Not a problem :)

MarkPieszak:
At least we can narrow down issues and what their regarding a little easier !

SteveSandersonMS:
Cool, thanks - closing as per conversation.

MarkPieszak:
Did you want me to remove the PULL_REQUEST template, or open a new PR with just the Issue template? @SteveSandersonMS

SteveSandersonMS:
Ah, I see. There's been some communication issues in this thread because I failed to be clear! I hadn't grasped that the "pull request" template and "issue" template were two different things and was referring to them both as a single thing.

What I've been trying to say is that I'd rather not add any of these templates.

While I'd be open to adding either/both of these *if it became unavoidable*, I think that at present, we're managing perfectly OK without them. From my point of view, adding mandatory form(s) for people to fill out creates a sense of bureaucracy and distances us from newcomers, so I'm keen to avoid it unless one day the project becomes unmanageable without them. Hope that's OK!

MarkPieszak:
Ohhh ok, it does you're right. Ok let's leave them out :) 
My fault I misunderstood too!

Issue #247, opened 2016-08-07
Can't find `aspnet-prerendering` when using a node version manager


aaronpowell:
I'm using a [node version manager PowerShell module](https://github.com/aaronpowell/ps-nvmw) (which I wrote) to allow me to have multiple versions of node installed and switch between them at will.

When using the SPA tooling, in particular the pre-rendering, there is an exception when a request is made:

```
fail: Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware[0]
      An unhandled exception has occurred: Call to Node module failed with error: To use prerendering, you must install the 'aspnet-prerendering' NPM package.
System.Exception: Call to Node module failed with error: To use prerendering, you must install the 'aspnet-prerendering' NPM package.
   at Microsoft.AspNetCore.NodeServices.HostingModels.HttpNodeInstance.<InvokeExportAsync>d__7`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.NodeServices.HostingModels.OutOfProcessNodeInstance.<InvokeExportAsync>d__12`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Microsoft.AspNetCore.NodeServices.NodeServicesImpl.<InvokeExportWithPossibleRetryAsync>d__9`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[...]

PR #248, opened 2016-08-07
FIxing path separator to address #247

aaronpowell:
When investigating #247 I noticed that the `NODE_PATH` would look something like:

`C:\tmp\.nvm\v6.3.1\node_modules:C:\_code\project\node_modules`

The problem is that the path separator is `:` not `;`.

This fix have resolve the issue in my testing.

## Testing fix

1. Create a new project using the yeoman generator
2. Enable server side rendering
3. Set `NODE_PATH` to have a value (eg: `C:\tmp`)
4. Run the application and experience the error described in #247 
5. Apply patch
6. Run the application successfully

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/HostingModels/OutOfProcessNodeInstance.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/HostingModels/OutOfProcessNodeInstance.cs
@@ -126,7 +126,7 @@ If you haven't yet installed node-inspector, you can do so as follows:
             var existingNodePath = Environment.GetEnvironmentVariable("NODE_PATH") ?? string.Empty;
             if (existingNodePath != string.Empty)
             {
-                existingNodePath += ":";
+                existingNodePath += ";";
             }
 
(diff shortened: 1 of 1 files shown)


aaronpowell:
Hmm just noticed that `:` is what *nix seems to use as a path separator (testing using Bash on Ubuntu on Windows). My primary dev environment is Windows.

aaronpowell:
Fixing to use `Path.PathSeparator` like it should do to pick `;` or `:`.

SteveSandersonMS:
Fantastic! Thanks @aaronpowell for tracking this down and preparing a nice clean fix :) It's really great to have this one resolved.

aaronpowell:
@SteveSandersonMS what's the release model for this? Wanting to pull it into some samples 😉

SteveSandersonMS:
I've pushed an updated set of beta packages (NodeServices, SpaServices, etc., all versioned `1.0.0-beta-000010`), so your fix should be present in any new projects you create, or in existing ones if you force it to update the NuGet packages. Thanks again for the PR!

PR #267, opened 2016-08-17
Bug(KnockoutSpa): Router TS issue

MarkPieszak:
TS erroring out (unable to build) due to `crossroads.normalizeFn = crossroads.NORM_AS_OBJECT;` <-- left hand assignment

```
ERROR in ./ClientApp/router.ts
(21,9): error TS2450: Left-hand side of assignment expression cannot be a constant 
or a read-only property.
```

### DIFF
--- a/templates/KnockoutSpa/ClientApp/router.ts
+++ b/templates/KnockoutSpa/ClientApp/router.ts
@@ -1,5 +1,5 @@
 import * as ko from 'knockout';
-import * as crossroads from 'crossroads';
+var crossroads = require('crossroads');
 
 // This module configures crossroads.js, a routing library. If you prefer, you
 // can use any other routing library (or none at all) as Knockout is designed to
--- a/templates/KnockoutSpa/ClientApp/router.ts
+++ b/templates/KnockoutSpa/ClientApp/router.ts
@@ -1,5 +1,5 @@
 import * as ko from 'knockout';
-var crossroads = require('crossroads');
+import crossroads = require('crossroads');
 
 // This module configures crossroads.js, a routing library. If you prefer, you
 // can use any other routing library (or none at all) as Knockout is designed to


SteveSandersonMS:
Thanks for reporting this.

I'm not following why TypeScript 2 is reporting that error. Is it a bug in TypeScript 2? Why does it think the LHS of the assignment is a constant?

If we do need to work around it, could we do so in a more targeted way that doesn't discard all type information for the `crossroads` package?

MarkPieszak:
It looks like changing this to `import crossroads = require('crossroads');` is the best bet, I meant to change it to that, apologies! That way we keep the intellisense.

Let me look into this more and see why it's creeping up now in the first place, I'll get back to you!

SteveSandersonMS:
Thanks. Looks like the reason is as discussed here: http://stackoverflow.com/questions/38818475/left-hand-side-of-assignment-expression-cannot-be-a-constant-or-a-read-only-prop

SteveSandersonMS:
Merged

MarkPieszak:
Since you were wondering as well, found out why this was creeping up now with 2.0. Apparently it was a TS bug because it *wasn't* showing up. This was fixed in 2.0.

http://stackoverflow.com/questions/38818475/left-hand-side-of-assignment-expression-cannot-be-a-constant-or-a-read-only-prop

This is because in es6 all module's variables are considered constants.

https://github.com/Microsoft/TypeScript/issues/6751#issuecomment-177114001
In TypeScript 2.0 the bug (of not reporting this error) was fixed.

Another workaround is that we could do `(crossroads as any).normalizeFn` ? Interesting.

> Since mongoose is still using the commonjs - var mongoose = require("mongoose") - not the es6 import syntax (which is used in the typings), you can suppress the error by assuming the module is of type any.
> 
> WORKAROUND:
> 
> (mongoose as any).Promise = global.Promise;

Issue #280, opened 2016-08-26
Using redirection during prerendering should be supported


ehaskins:
I'm implementing some client side-authentication logic in React while using the asp-prerender. Client side this this isn't a problem using react-router's onEnter hook, but when prerender it needs to return 302. 

The react-router [ServerRendering](https://github.com/reactjs/react-router/blob/master/docs/guides/ServerRendering.md) page shows how to determine if a redirect should occur before doing a complete render, but I don't see any way to actually initiate the redirect. 

Is it possible to initiate a 302 from a TagHelper?

For now I can either see two workarounds
 - Disable pre-rendering, and allow the navigation to happen client side
 - Use asp.net routing to redirect before attempting to do the 302, but this seems fragile.

I suspect enabling this is a larger task, but the current error message in the React-Redux template is `Exception: Call to Node module failed with error: Error: The location '/' doesn't match any route configured in react-router.`, which isn't accurate. A more descriptive message would have been very helpful.

SteveSandersonMS:
Great suggestion - thanks for this!

> Is it possible to initiate a 302 from a TagHelper?

Thankfully, yes it is, so I've just added support for redirections :)

To get this feature in your project, update to the latest `aspnet-prerendering` NPM module (1.0.6 or later). Also you'll need to update to a newer version of `Microsoft.AspNetCore.SpaServices`, but I haven't yet published a version that includes this. You can build your own from source if you need, or wait a little until we drop an update.

Then your boot script can return a `redirectUrl` and the server will perform it. You can see this at work in the updated `boot-server.ts` file for the ReactRedux template: https://github.com/aspnet/JavaScriptServices/commit/28550784edb6c9c453a8983285dc0d6a9aaa12a9

ehaskins:
Awesome! I'll be keeping an eye out for the next update.

SteveSandersonMS:
It's published now

PR #281, opened 2016-08-26
Improve prerender error message

ehaskins:
Add a more specific error message which would have saved me some time. 

Addresses #280.


SteveSandersonMS:
Thanks for contributing this! Since support for redirections has now been added (see #280), I think we no longer need this improved error message since the redirection would now be handled properly.

PR #319, opened 2016-09-17
ng2 2.0, Universal 2.0, TS 2.0, Preboot 4.*

MarkPieszak:
Angular2.0 updates (NgModule, Router-v3, etc)
Universal 2.*
Preboot 4.*
TypeScript 2.0 (needed for latest Universal)

Tested with JavaScript enabled & disabled.

PR #321, opened 2016-09-19
Update Angular2Spa template to Angular 2.0/TypeScript 2.0, based on @MarkPieszak's PR

SteveSandersonMS:
Following on from @MarkPieszak's PR #319, I've simplified the template further as far as I think makes sense at the moment.

Putting this out as a PR for feedback before merging and publishing.


SteveSandersonMS:
@MarkPieszak, do you have any feedback on the simplifications/changes I've made here? Your PR was extremely helpful. On top of that, the main things I wanted to change were:

* Making HMR work
* Removing files/folders and NPM dependencies that we can avoid the need for
* Adjusting the filename/folder conventions to more closely match those emitted by angular-cli
* Simplifying boot code (just because few developers are going to understand this at first glance)
* Simplifying webpack config (eliminating dev/prod override files). I have plans to simplify this further, but will do that when I get time to update all the templates, not just the Angular one.

I also attempted to get the Angular 2 compiler (`ngc`), or at least Webpack 2 tree shaking, to work with this, but completely failed. It doesn't seem like `ngc` and arbitrary Webpack configs work together right now, so that will have to wait, unless someone else has a way to make it work! There are a bunch of experimental webpack-ngc-loader type things on GitHub right now, but it's not clear that any of them really work or will be supported in the long run.

@hheexx - check the later commits. It's in `dependencies` because it's needed at runtime if you're doing server-side prerendering.

templates/Angular2Spa/ClientApp/boot-server.ts
```diff
@@ -1,76 +1,28 @@
...
-        return { html };
-
-    }).catch(err => {
-        
-        console.log(err);
-        return { html : doc };
-
     });
 
+    return requestZone.run<Promise<string>>(() => platform.serializeModule(AppModule))
```
MarkPieszak: Much cleaner I like it :+1:

templates/Angular2Spa/ClientApp/app/app.module.ts
```diff
@@ -0,0 +1,31 @@
...
+    bootstrap: [ AppComponent ],
+    declarations: [
+        AppComponent,
+        NavMenuComponent,
+        CounterComponent,
+        FetchDataComponent,
+        HomeComponent
+    ],
+    imports: [
+        UniversalModule,
```
MarkPieszak: Can we leave a Note that `UniversalModule` should be the first Import?\

Also, we should point out that `BrowserModule` `HttpModule` `JsonpModule` are all **automatically included** here.
(The same goes for the Node versions on the main.server.ts side, NodeModule / NodeHttpModule etc)
Things can error out if they include those. We're working on looking into fixing that.

Patrick also thought it'd be nice to keep the main files separate since that allows someone to have specific platform logic injected for either browser/server, but it's up to you. It's definitely cleaner/simpler this way!

hheexx:
Oh.. I see they are removed in later commit.

templates/Angular2Spa/ClientApp/app/app.module.ts
hheexx: Please keep server and client main files separate.
For me they diverged very early in development. And I am pretty sure it will for most users.

templates/Angular2Spa/ClientApp/boot-server.ts
```diff
@@ -1,76 +1,28 @@
...
-        originUrl: params.origin
-    };
-
-    // defaults
-    var cancel = false;
-
-    const _config = Object.assign({
-      get cancel() { return cancel; },
-      cancelHandler() { return Zone.current.get('cancel') }
-    }, platformConfig);
```
SteveSandersonMS: OK, please let me know if you think something needs to be added for this.

templates/Angular2Spa/ClientApp/app/app.module.ts
SteveSandersonMS: Added the comment about `UniversalModule` (and BrowserModule/HttpModule/JsonpModule)

>  main files separate
I'd hope that the boot files serve that purpose well enough (this is where you configure DI, after all), and am reluctant to add more duplicate files. Having one root `@NgModule` makes things much clearer, especially given how much tedious config goes into that file. We can monitor this and see if people often report issues that are best solved by having separate root modules.

templates/Angular2Spa/ClientApp/main.browser.ts
```diff
@@ -1,55 +0,0 @@
...
-  // Our Components
-  declarations: [
-    App, Counter, FetchData, Home, NavMenu 
-  ],
-
-  imports: [
-    
-    // * NOTE: Needs to be your first import (!)
-    UniversalModule, 
-    // * ^ BrowserModule, HttpModule, and JsonpModule are included here
```
SteveSandersonMS: Done

SteveSandersonMS:
> One thing recommended by the Angular team is DOMContentLoaded

Thanks for the feedback. Done.

> As for AOT Compilation, right now we don't have it fully working, it would be additional code needed in the boot-server, but there's more that needs to be done to get that working on our end.

The issue isn't just making it work at runtime in Univeral. Right now I'm not successful in being able to actually do the AOT compilation at all, because it's not generally compatible with arbitrary webpack configs (e.g., using `raw-loader` and `require` to reference template `.html` files).

> Webpack2 does work with Universal, what problems were you getting with it

No, that part works fine. I was meaning to say that the end-to-end story with Webpack 2 tree shaking isn't usable right now because of the inability to precompile Angular 2 templates.

templates/Angular2Spa/ClientApp/app/app.module.ts
hheexx: Can we have logic in main file based on platform? Then we don't need 2 files.

if (server) {
declarations: [......]
} else if (client) {
declarations: [.....]
}

?

SteveSandersonMS: @hheexx Yes, just `import { isNode, isBrowser } from 'angular2-universal';` and then do `if (isNode) { ... }` etc.

SteveSandersonMS:
Thanks for the feedback! This is now merged into the default branch. I'll publish actual Yeoman templates tomorrow.

templates/Angular2Spa/ClientApp/app/app.module.ts
hheexx: That is sweet. I would add it as a comment so people know.

MarkPieszak:
Excellent! :)
If we need to circle back and add the cancelHandler stuff later we can take care of it, for now at least this should be good enough! I'm still looking into it how important it is, and if it's possible for us to handle it internally with Universal instead perhaps.

Cheers.

PR #351, opened 2016-10-03
Update Dockerfile

korneel:
The app doesn't run in Docker on base image `microsoft/dotnet:latest` because Microsoft.NETCore.App 1.0.1 isn't installed. Only 1.0.0 is installed. The app runs fine on base image `microsoft/dotnet:1.0.0-preview2-onbuild`.


SteveSandersonMS:
Thanks for this!

PR #360, opened 2016-10-05
Templates: (JS/TS) change ' to "

stephtr:
I converted all quotes to double qoutes in ClientApp and webpack.config.* files


MarkPieszak:
You're removing the placeholder text files that I believe Steve needs for the generators and the old tsd files that are currently necessary for the project until it's moved to TS2.

stephtr:
Oh sorry, I didn't want these two commits to be included into the pull request.

MarkPieszak:
No worries just letting you know! :)
I'm about to check and make sure the TS2 PR is all set with ReactRedux template and then all of them will have that stuff removed and updated as well!

stephtr:
Fixed it.

SteveSandersonMS:
First, thanks @stephtr for submitting this. I appreciate the effort you've taken!

At the moment, I'm not sold on the idea of changing to `"`. I know what it says in TypeScript's [coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines), but:

1. That guide tries to say that some parts of it are only "*meant for contributors to the TypeScript project*", whereas other parts "*apply as to how idiomatic TypeScript code should be written (e.g. the "Names" and "Style" sections)*" (which the "quotes" part is not in)
2. Much more importantly (to such a degree that it renders item 1 almost irrelevant anyway), the JavaScript community and in turn the TypeScript community is nearly unanimous in using single quotes. All of [Angular 2 is coded this way](https://github.com/angular/angular/blob/master/modules/%40angular/core/src/application_module.ts), and they [document it that way](https://angular.io/docs/ts/latest/quickstart.html), and their popular [starter kits are the same](https://github.com/mgechev/angular2-seed/blob/master/src/client/app/app.component.ts). The [official Angular 2 CLI](https://github.com/angular/angular-cli) also emits projects that use `'`.
3. Among React apps, there isn't a single most-obvious authoritative TypeScript example, but all of the first few starter kits I've found ([this](https://github.com/alex3165/react-typescript-starter/blob/master/src/App.tsx), [this](https://github.com/coryhouse/react-typescript-starter-kit/blob/master/src/Goal.tsx), [this](https://github.com/tomastrajan/react-typescript-webpack/blob/master/src/main.app.service.ts)) use `'`. I didn't find any that use `"` (though I stopped looking after those first three). Interestingly, React's official starter kit (linked from [their main docs here](https://facebook.github.io/react/docs/getting-started.html)), which uses JS not TS, consistently uses `'` in JS code and `"` around HTML attribute values.

Considering the degree of consistency in the community on preferring `'`, it seems like the TypeScript team's use of `"` in their own source code is more of a historical quirk, and if we used it, we'd look extremely out-of-step with the rest of the JS/TS community.

Now, as we all know, code style discussions (like naming discussions) tend to be extremely contentious and emotional. So at this point I don't really think it's a great idea to have a general discussion about people's preferences - it will just be frustrating to everyone. Instead I'm going to take the risk of closing this so we can judge in another month or so whether other people keep bringing it up. I know that will be frustrating to @stephtr so I apologise for that and thank you again for doing the work to create this PR. Let's see if we can stay consistent with the broader community rather than having yet another awkward cultural gap between Microsoft-focused devs and everyone else. Hope that's OK for now!

Issue #159, opened 2016-07-01
IIS Deployment Error - Call to Node module failed with error: Error: no such file or directory at MemoryFileSystem.readFileSync


chris-herring:
I can run the project (React & Redux template) fine using dotnet run but after running "dotnet publish -c Release" and then hosting the build in IIS I hit the below exception. Anyone had success with this? I've got the latest code, set the IIS AppPool to unmanaged code and given the app pool user full control over the entire directory. 

```
Exception: Call to Node module failed with error: Error: no such file or directory
at MemoryFileSystem.readFileSync (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\memory-fs\lib\MemoryFileSystem.js:114:10)
at C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\aspnet-webpack\LoadViaWebpack.js:70:65
at Compiler.<anonymous> (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:194:14)
at Compiler.emitRecords (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:282:37)
at Compiler.<anonymous> (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:187:11)
at C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:275:11
at Compiler.applyPluginsAsync (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\tapable\lib\Tapable.js:60:69)
at Compiler.afterEmit (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:272:8)
at Compiler.<anonymous> (C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\webpack\lib\Compiler.js:267:14)
at C:\Development\Test Projects\CoreReactTest\bin\Release\netcoreapp1.0\publish\node_modules\async\lib\async.js:52:16
```

chris-herring:
[...]

PR #362, opened 2016-10-05
Update all projects to TypeScript 2.0 - remove typings - etc

MarkPieszak:
All projects up & running now with TS2.0.


chokidar/package.json
```diff
@@ -0,0 +1,123 @@
...
+      {
+        "raw": "chokidar@^1.0.0",
+        "scope": null,
+        "escapedName": "chokidar",
+        "name": "chokidar",
+        "rawSpec": "^1.0.0",
+        "spec": ">=1.0.0 <2.0.0",
+        "type": "range"
+      },
+      "C:\\Users\\mark.pieszak\\Documents\\GitHub\\JavaScriptServices\\templates\\KnockoutSpa\\node_modules\\watchpack"
```
tinchou: This is probably not right (your path). But what is chokidar doing here anyway?

MarkPieszak: I'm not sure removing it now... just realized that too

SteveSandersonMS:
Thanks for the update @MarkPieszak!

There seems to be some unrelated stuff in the PR that was perhaps added by mistake. For example, the `chokidar` source code for some reason, and renaming `boot-client.ts` to `main-client.ts` etc. 

Could you remove the unrelated changes from this PR so that it's just the updates for TS 2.0?

templates/KnockoutSpa/ClientApp/boot.ts
```diff
@@ -1,8 +1,8 @@
 import './css/site.css';
 import * as ko from 'knockout';
-import { createHistory } from 'history';
 import './webpack-component-loader';
 import AppRootComponent from './components/app-root/app-root';
+var createHistory = require('history').createBrowserHistory
```
SteveSandersonMS: Is this change deliberate? Could you clarify why it would be needed (hopefully it isn't), and if it is, change the `var` to `const` and add the semicolon at the end of the line?

templates/KnockoutSpa/tsconfig.json
```diff
@@ -3,8 +3,13 @@
...
     "target": "es5",
     "sourceMap": true,
-    "skipDefaultLibCheck": true
+    "skipDefaultLibCheck": true,
+    "typeRoots": [ "node_modules/@types" ],
+    "types": ["core-js", "crossroads", "history", "knockout",
+     "requirejs", "signals", "whatwg-fetch" ]
   },
+  "compileOnSave": false,
+  "buildOnSave": false,
```
SteveSandersonMS: Are these two lines really needed?

templates/ReactReduxSpa/ClientApp/configureStore.ts
```diff
@@ -1,27 +1,29 @@
-import { createStore, applyMiddleware, compose, combineReducers } from 'redux';
-import * as thunkModule from 'redux-thunk';
-import { routerReducer } from 'react-router-redux';
-import * as Store from './store';
-import { typedToPlain } from 'redux-typed';
+import * as Redux from "redux";
```
SteveSandersonMS: For consistency, can we go back to single-quotes for all lines in this file?

templates/ReactReduxSpa/ClientApp/configureStore.ts
```diff
@@ -1,27 +1,29 @@
...
+import { createStore, applyMiddleware, compose, combineReducers, GenericStoreEnhancer } from "redux";
+import thunk from "redux-thunk";
+import { routerReducer } from "react-router-redux";
+import * as Store from "./store";
+import { typedToPlain } from "redux-typed";
+
```
SteveSandersonMS: There's already a blank line below this - don't need another :)

templates/ReactReduxSpa/ClientApp/configureStore.ts
MarkPieszak: Sure!

templates/ReactReduxSpa/ClientApp/configureStore.ts
MarkPieszak: Haha sounds good, be right back (grabbing lunch) and I'll fix the above comments :+1:

templates/ReactReduxSpa/tsconfig.json
```diff
@@ -5,7 +5,12 @@
     "jsx": "preserve",
     "experimentalDecorators": true,
     "sourceMap": true,
-    "skipDefaultLibCheck": true
+    "skipDefaultLibCheck": true,
+    "lib": ["es6", "dom"],
+    "typeRoots": [ "node_modules/@types" ],
```
SteveSandersonMS: Is this needed?

templates/KnockoutSpa/ClientApp/components/app-root/app-root.ts
```diff
@@ -40,4 +40,4 @@ class AppRootViewModel {
     }
 }
 
-export default { viewModel: AppRootViewModel, template: require('./app-root.html') };
+export default { viewModel: AppRootViewModel, componentName: 'Home', template: require('./app-root.html') };
```
MarkPieszak: This might of been accidental, Knockout is giving me a minor :
Any idea what that might be? @SteveSandersonMS 

```vendor.js?v=P985CJI4_0EhN8wTjSzDXZnMQt78MgU91Uz14XJKMk8:26089 
Uncaught Error: Unable to process binding 
"component: function (){return { name:route().page,params:route} }"
Message: No component name specified```

File: chokidar/package.json (hunk 1 above)
Comment (MarkPieszak) [inline chokidar/package.json]: Removed it now!

File: templates/KnockoutSpa/ClientApp/boot.ts (hunk 2 above)
Comment (MarkPieszak) [inline templates/KnockoutSpa/ClientApp/boot.ts]: Be back shortly and i'll talk to you about this one.

File: templates/KnockoutSpa/ClientApp/components/app-root/app-root.ts (hunk 7 above)
Comment (SteveSandersonMS) [inline templates/KnockoutSpa/ClientApp/components/app-root/app-root.ts]: Sounds like the URL isn't matching any route, so `route()` is an empty object, so `route().page` is undefined, so you're returning nothing for the `name` of the component.

File: templates/Angular2Spa/webpack.config.js [hunk 8]
```diff
@@ -41,7 +41,7 @@ var clientBundleConfig = merge(sharedConfig, {
 
 // Configuration for server-side (prerendering) bundle suitable for running in Node
 var serverBundleConfig = merge(sharedConfig, {
-    entry: { 'main-server': './ClientApp/boot-server.ts' },
+    entry: { 'main-server': './ClientApp/main-server.ts' },
```
SteveSandersonMS: Let's keep renaming separate from this PR. Can you revert the names?

templates/ReactReduxSpa/tsconfig.json
```diff
@@ -5,7 +5,12 @@
...
+    "types": [ "react", "react-dom", "react-redux", "react-router", "react-router-redux",
+     "redux", "redux-thunk", "source-map", "uglify-js", "webpack", "webpack/webpack-env",
```
stephtr: Even though it is not yet at npm: `webpack-env` has been outsourced into an additional package ([DefinitelyTyped #11684](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11684)).

templates/ReactReduxSpa/ClientApp/configureStore.ts
```diff
@@ -1,27 +1,29 @@
...
 
 export default function configureStore(initialState?: Store.ApplicationState) {
     // Build middleware. These are functions that can process the actions before they reach the store.
-    const thunk = (thunkModule as any).default; // Workaround for TypeScript not importing thunk module as expected
-    const windowIfDefined = typeof window === 'undefined' ? null : window as any;
-    const devToolsExtension = windowIfDefined && windowIfDefined.devToolsExtension; // If devTools is installed, connect to it
+    const windowIfDefined = typeof window === "undefined" ? null : window as any;
+    // If devTools is installed, connect to it
+    const devToolsExtension = windowIfDefined && windowIfDefined.devToolsExtension as () => GenericStoreEnhancer;
     const createStoreWithMiddleware = compose(
```
stephtr: Not relevant yet, but I wanted to note it somewhere: This line won't work with the most recent version of redux.d.ts ([redux #1936](https://github.com/reactjs/redux/pull/1936#issuecomment-250443020)) and propably have to be changed in the future to `compose<Redux.StoreEnhancerStoreCreator<Store.ApplicationState>>(`, which doesn't work with the version available now on npm.

templates/ReactReduxSpa/ClientApp/configureStore.ts
```diff
@@ -1,27 +1,29 @@
-import { createStore, applyMiddleware, compose, combineReducers } from 'redux';
-import * as thunkModule from 'redux-thunk';
-import { routerReducer } from 'react-router-redux';
-import * as Store from './store';
-import { typedToPlain } from 'redux-typed';
+import * as Redux from "redux";
+import { createStore, applyMiddleware, compose, combineReducers, GenericStoreEnhancer } from "redux";
```
MarkPieszak: Just noticed that, somehow overlooked that line too! :P

templates/Angular2Spa/webpack.config.js
MarkPieszak: It was breaking since webpack is looking for a file by that name, want me to change it there instead?

templates/KnockoutSpa/tsconfig.json
MarkPieszak: They are just helpful depending on certain IDE's to disable automatic `tsc` compilation / etc.
Do you want me to remove them? Just thought they were helpful to ensure webpack handles everything.

templates/KnockoutSpa/ClientApp/components/app-root/app-root.ts
```diff
@@ -4,7 +4,7 @@ import navMenu from '../nav-menu/nav-menu';
 
 // Declare the client-side routing configuration
 const routes: Route[] = [
-    { url: '',              params: { page: 'home-page' } },
+    { url: '/',              params: { page: 'home-page' } },
```
MarkPieszak: This was an attempt to fix a knockout issue, I can revert this change!

templates/ReactReduxSpa/tsconfig.json
stephtr: Theoretically it should be already online. I filed an issue at [types-publisher #159](https://github.com/Microsoft/types-publisher/issues/159).

templates/ReactReduxSpa/tsconfig.json
MarkPieszak: Not specifically, it's just nice to see at a glance what Type definitions are being used.

templates/KnockoutSpa/ClientApp/components/app-root/app-root.ts
MarkPieszak: So when a refresh is done it doesn't load that routed content, but when the link is clicked, it works fine. Any idea what could be causing this?

templates/KnockoutSpa/ClientApp/boot.ts
MarkPieszak: So it needs to be updated because when the previous `import {}` way was done, with the new History, it doesn't call the correct methods and ends up erroring that "createBrowserHistory" doesn't exist. I've tried the other ways suggested on the packages readme (https://github.com/mjackson/history) but only this one works). Updating it to `const` now.

templates/KnockoutSpa/tsconfig.json
```diff
@@ -3,8 +3,13 @@
     "moduleResolution": "node",
     "target": "es5",
     "sourceMap": true,
-    "skipDefaultLibCheck": true
+    "skipDefaultLibCheck": true,
+    "typeRoots": [ "node_modules/@types" ],
```
MarkPieszak: Removed the `typeRoots` part: (`node_modules/@types`) is default.
The part where `types:[]` is set is needed otherwise it won't make them necessary during compilation.

templates/ReactReduxSpa/ClientApp/configureStore.ts
MarkPieszak: Fixed.

MarkPieszak:
@SteveSandersonMS This should be all set now, minus the content not loading right away in the KnockoutSpa template (maybe you know what's causing it?)

Removed the typeRoots, changed createHistry to `const`
Eliminated double Enters, Double quotes on the one page, etc.

stephtr:
Just for information: The redux-typed/StrongActions.d.ts error is still remaining.

templates/KnockoutSpa/tsconfig.json
stephtr: Why are they only in one template and not all?

templates/KnockoutSpa/ClientApp/boot.ts
stephtr: It seems to be connected to tsconfig.json "target": "es5". Switching to es6 results in an error with crossroads.

MarkPieszak:
Yes, strangely the error comes from within node_modules itself. Made an issue, but seems no one has responded to it, also doesn't seem to prevent anything from working luckily!

templates/ReactReduxSpa/tsconfig.json
stephtr: @types/webpack-env is online now.

templates/ReactReduxSpa/tsconfig.json
```diff
@@ -5,7 +5,11 @@
...
     "experimentalDecorators": true,
     "sourceMap": true,
-    "skipDefaultLibCheck": true
+    "skipDefaultLibCheck": true,
+    "lib": ["es6", "dom"],
+    "types": [ "react", "react-dom", "react-redux", "react-router", "react-router-redux",
+     "redux", "redux-thunk", "source-map", "uglify-js", "webpack", "webpack/webpack-env",
+      "whatwg-fetch" ]
   },
   "exclude": [
```
antmdvs: @SteveSandersonMS Should `wwwroot` be excluded as well? If so, can we sneak that in this PR?

templates/ReactReduxSpa/ClientApp/configureStore.ts
stephtr: Just another note on these lines, they could be replaced by (like in the [redux documentation](http://redux.js.org/docs/api/compose.html)):
```
const store = createStore(
	buildRootReducer(Store.reducers),
	initialState,
	compose(
		applyMiddleware(thunk, typedToPlain),
		devToolsExtension ? devToolsExtension() : f => f
	)
);
```
Which also eliminates one typecast.

templates/ReactReduxSpa/tsconfig.json
MarkPieszak: great, i'll replace `webpack/webpack-env` in here with that `@types/webpack-env`. Thanks @stephtr !

templates/ReactReduxSpa/ClientApp/configureStore.ts
stephtr: `compose(...) as StoreEnhancer<Store.ApplicationState>` would be an option which is compatible with redux 3.6.0` **and** upcoming versions.

stephtr:
As far as I noticed there is a `declare var module: any;` in `ReactReduxSpa/ClientApp/routes.tsx`, which isn't needed anymore.

templates/ReactReduxSpa/tsconfig.json
antmdvs: I defer to @SteveSandersonMS, as I'm not even sure if it's necessary. Just something I noticed.

MarkPieszak:
@SteveSandersonMS Everything has been updated to the latest from master.
The only issue remaining (as far as I've seen) is KnockoutSpa not painting that initial Route if you could take a look?

I didn't want to add angular2 recent upgrades (2.1 & universal 2.1 patches) until we were finished with this one yet. Let me know what you think!

emzero:
The `StrongActions.d.ts` error is easily fixed by setting the `dispatch` argument type to `Dispatch<TState>`, but since Redux now includes its own typings, wouldn't be better to leverage them so we stay as close to "official" as possible? 
Also, that `redux-typed` package seems quite unpopular.

MarkPieszak:
Great idea, totally agree, rather it be as official as possible.
@radu-porumb em zeros comment above looks like what we should be doing.

radu-porumb:
@MarkPieszak That's exactly what I did to move things along. But my project is on a CI pipeline which means as soon as it gets to source control the build server pulls it, builds it and fails spectacularly. I could always just throw out `redux-typed` and copy-paste the code with the `<TState>` fix into a local file to keep everything strongly typed but I'd like to know if we're likely to get a package update soon before going that route.

templates/KnockoutSpa/tsconfig.json
SteveSandersonMS: Have removed them

SteveSandersonMS:
@stephtr:
> As far as I noticed there is a declare var module: any; in ReactReduxSpa/ClientApp/routes.tsx, which isn't needed anymore.

Well spotted! Removed.

> Just for information: The redux-typed/StrongActions.d.ts error is still remaining.

Thanks - fixed (in newly-published `redux-typed` version 1.0.1).

@MarkPieszak:
> The only issue remaining (as far as I've seen) is KnockoutSpa not painting that initial Route if you could take a look?

The issue was that the `history` module had a breaking change between versions 2 and 4 whereby it no longer fires a navigation event on page load. I've amended the code to populate the initial history state. Unfortunately `@types/history` is massively out-of-date (it still describes the API for `history` version 2) so there's another cast-to-`any` to make this compile.

> I didn't want to add angular2 recent upgrades (2.1 & universal 2.1 patches) until we were finished with this one yet. Let me know what you think!

Makes sense. I'm merging this PR now so let's look at 2.1 soon :) Thanks.

@emzero:
> Redux now includes its own typings, wouldn't be better to leverage them so we stay as close to "official" as possible? 

Mostly I agree with you but it's a wider discussion, so let's move this part of the discussion to a separate issue.

templates/ReactReduxSpa/tsconfig.json
SteveSandersonMS: @antmdvs I'm not sure why `wwwroot` should be excluded here, but if you think there's something wrong, could you please file a separate issue specifying what problem occurs and what should be done to fix it? Thanks!

SteveSandersonMS:
Merged! Thanks @MarkPieszak for your work on this, and for accounting for many changes while waiting for it to be merged.

MarkPieszak:
Not a problem, glad to be a part of it all :)
Let me know if there's any other random issues you'd like me to look into? @SteveSandersonMS

SteveSandersonMS:
> Let me know if there's any other random issues you'd like me to look into

@MarkPieszak The biggest thing for the Angular2Spa template would be if there was some way to get AoT compilation and tree-shaking. But it would have to be pretty much transparent to the developer and not make the webpack config a lot more complex. I don't know if that's possible yet, because last time I checked, `ngc` didn't have any supported way to work in a Webpack environment.

tinchou:
It's transparent with @ngtools/webpack. But I have to admit I don't see a
great size improvement with tree shaking. Maybe I'm missing something.

I also mentioned I don't know if it'll work with Universal because I
removed it from my project.

I can share my webpack config later if you want!

MarkPieszak:
Tinchou is right, ngtools/webpack is avaialable, but there's still a lot of things in the works to make everything more stable.

On our end, on the Universal side, we still have a little bit of ground to cover, there has been some work done to get some basic AoT going, but I think it might be a good idea to let things mature a little more before moving it over here as well? (Angular Core itself needs to mature AoT itself in a few areas)
The performance/speed improvements (so far) are incredible, so it would be amazing to get it in here :+1: 

But that sounds good to me! I'll start playing around with some things, at least open up a PR sometime so we can begin looking at webpack2's tree-shaking & AoT combination for the template.

tinchou:
Here's my example configuration https://gist.github.com/tinchou/08d1be519b525899969180bdf7edb67f

PR #452, opened 2016-11-21
Generate documentation using compodoc

gsaadeh:
Generate static HTML docs for the Angular2 application using [compodoc](https://github.com/compodoc/compodoc)

### DIFF
--- a/templates/Angular2Spa/package.json
+++ b/templates/Angular2Spa/package.json
@@ -2,7 +2,9 @@
   "name": "Angular2Spa",
   "version": "0.0.0",
   "scripts": {
-    "test": "karma start ClientApp/test/karma.conf.js"
+    "test": "karma start ClientApp/test/karma.conf.js",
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for the suggestion! I wasn't familiar with compodoc before, so this is interesting to see.

For the time being, I think it would be better if we didn't include this by default, since we try as much as possible to minimise the number of concepts and technologies that we push onto developers. If this becomes a widely requested feature in the future, we can definitely reconsider that, but so far I haven't heard anyone else asking for this.

Is that OK? I'll close for now but further comments are welcome.

gsaadeh:
Hi Steve,

Thanks for your response. Generating documentation is widely used in Angular2 applications including Angular Seed and Advanced Angular Seed and both of these two templates are using Compodoc. The Webpack Starter template from AngularClass also has it but it's using the more traditional [TypeDoc](http://typedoc.org).

If you get any request for it, you can add it back.

Cheers,

George

PR #497, opened 2016-12-07
Fix filename in sample

sebastienros:
Mismatch between code sample and described js filename.


SteveSandersonMS:
Thanks @sebastienros!

Issue #518, opened 2016-12-16
Invoke module with unknown return type


nicemd:
Hi,
I'd like to invoke a node module that I on forehand don't know if it will return a stream or a json result. If I call `InvokeAsync<object>()` it works for json but for streams it throws the exception:

_System.ArgumentException: Node module responded with binary stream. This cannot be converted to the requested generic type: System.Object.Instead you must use the generic type System.IO.Stream._

But obviously I can't use `InvokeAsync<Stream>()` if the result is json.

```            
        public static async Task MainAsync(string[] args)
        {
            var services = new ServiceCollection();
            services.AddNodeServices(options => {
                options.ProjectPath = Directory.GetCurrentDirectory();
            });
    
            var serviceProvider = services.BuildServiceProvider();
            var nodeServices = serviceProvider.GetRequiredService<INodeServices>();

            var result = await nodeServices.InvokeExportAsync<object>("test.js", "jsonfunc"); // works
            result = await nodeServices.InvokeExportAsync<Stream>("test.js", "streamfunc"); // works
            result = await nodeServices.InvokeExportAsync<object>("test.js", "streamfunc");  // throws
            if (result is Stream)
            {
                // read from stream
            }
            else if (result is JObject)
            {
                // handle json response
            }
        }
```

test.js
```
module.exports.jsonfunc = function (callback) {

    callback(null, { "hello": "world" });
};

module.exports.streamfunc = function (result) {
    result.stream.write("hello world");
    result.stream.end();
};
```

I'm using "Microsoft.AspNetCore.NodeServices": "1.1.0-beta-000002"

SteveSandersonMS:
Fixed with @nicemd's pull request #519.

For anyone reading, this now means that if you declare the generic type on your `InvokeAsync` as `object`, then the result can be either a `Stream` or some other object type depending on what your Node.js code chooses to return.

PR #519, opened 2016-12-16
Streamed responses can be be Object or Stream

nicemd:
See issue #518


SteveSandersonMS:
Thanks - this looks great!

Issue #534, opened 2016-12-22
Deploy to Azure Cannot find module 'angular2-universal-polyfills'


Gimly:
Not quite sure if this is an error related to the template or not, but I've created a test app with the aspnetcore-spa yo template with Angular 2 and tried deploying it to Azure App Services.

I get the following error:
```
Exception: Call to Node module failed with error: Prerendering failed because of error: Error: Cannot find module 'angular2-universal-polyfills'
at Function.Module._resolveFilename (module.js:469:15)
at Function.Module._load (module.js:417:25)
at Module.require (module.js:497:17)
at require (internal/module.js:20:19)
at Object.<anonymous> (D:\home\site\wwwroot\ClientApp\dist\main-server.js:87:19)
at __webpack_require__ (D:\home\site\wwwroot\ClientApp\dist\main-server.js:20:30)
at Object.<anonymous> (D:\home\site\wwwroot\ClientApp\dist\main-server.js:48:2)
at __webpack_require__ (D:\home\site\wwwroot\ClientApp\dist\main-server.js:20:30)
at D:\home\site\wwwroot\ClientApp\dist\main-server.js:40:18
at Object.<anonymous> (D:\home\site\wwwroot\ClientApp\dist\main-server.js:43:10)
Current directory is: D:\home\site\wwwroot
```
The code works fine in Visual Studio as well as locally and I'm not sure what I'm missing to make it work once published.

SteveSandersonMS:
That's strange. The publishing certainly works as far as I'm aware.

Is it possible you forgot to specify a Node version when setting up your Azure website? See: http://blog.stevensanderson.com/2016/10/04/angular2-template-for-visual-studio/#now-lets-deploy

The reason I suggest this is that, if you don't specify a Node version, then it falls back on an old version with NPM 2.x, which doesn't correctly restore all Angular NPM packages. That would at least be consistent with the error message you're getting.

Can you provide repro steps based on deploying to a completely new Azure website? (Don't try re-deploying to your broken one, as you might now have a corrupted `node_modules` directory that continues to break further deployments)

SteveSandersonMS:
I've been testing Azure deployment from several angles today, and didn't hit any issue like this. Since we don't have any information on how to reproduce it, I'm going to put it down to something unique to your scenario (e.g., a network failure during NPM restoration) and close this.

If you are able to describe how to reproduce the issue, please reopen with details!

Gimly:
[...]

PR #572, opened 2017-01-09
Updated all template project.json tools to use 1.1

Gimly:
I figured out what was causing the issue that I had opened in #534. It was not really linked to Angular or anything as the error message could suggest, but to the publishing process.

I'm publishing my app through Bitbucket Pipelines that runs  a script in a Docker image to create a CI/CD system. The script I was using was simply running `dotnet restore` and then `dotnet publish`. What I hadn't seen was that `dotnet publish-iis` that is called by the post publish methods was failing. It was failing because on the docker image I'm using (microsoft/aspnetcore-build) only the .Net Core 1.1 is installed. The issue was that there were still tools dependencies in the project.json file that were depending on .Net Core 1.0, and running those tools were making the publish-iis fail.

Using the latest 1.1.0-preview4-final versions of

- Microsoft.AspNetCore.Razor.Tools
- Microsoft.AspNetCore.Server.IISIntegration.Tools
- Microsoft.DotNet.Watcher.Tools

fixes the issue and makes the all publish work correctly on the docker machine.


Gimly:
Hello,

Is there something wrong with this pull request? Do you want me to fix something?

Regards

RichiCoder1:
According to #684 this might be rendered redundant. AspNet peeps are probably in fire mode with VS RTM coming down the pipeline. /cc @natemcmaster

SteveSandersonMS:
@Gimly Sorry this is taking a while to merge and for not commenting so far. Thanks for this contribution.

We're just working through transitioning everything in this repo to `csproj`-based tooling. Once that is done, some parts of this PR will still apply (the bits in the templates), and some parts will not (the bits in the samples). If we are retaining support for generating `project.json`-based projects, I'll then merge in the relevant parts of this PR. Hope that's OK!

Gimly:
@SteveSandersonMS No problem! I was just wondering if there was something I could improve in the PR that was blocking your from merging. I completely understand that you're merging to the csproj-based tooling and that it has an impact.

SteveSandersonMS:
Again, sorry this has taken a while to get to.

Now that dotnet's 1.0 `csproj` tooling has shipped, we'll shortly be pulling out the `project.json` vestiges from here. Obviously people who already have `project.json`-based projects can continue with them for as long as they want, but it doesn't really make sense for us to be shipping and maintaining tools to create *new* `project.json`-based projects. Newly-created projects should be `csproj`-based.

As such, the changes in this PR (which would have otherwise been valuable) will not be applicable any more. So to keep things tidy, I'll close this now.

Again, thanks for contributing this and for bearing with us while we make the transition to 1.0.0 tooling!

PR #615, opened 2017-01-27
Update templates to Webpack 2

SteveSandersonMS:
I'll merge this at the start of next week, but posting it here as a PR first just in case anyone can spot any issues.

Moving to `awesome-typescript-loader` improves the Angular2Spa Webpack build times (down from about 7.5s to about 4s on my machine). It makes no difference to the other templates, which were already fast to build, but I've migrated them to `awesome-typescript-loader` for consistency.

Anyone who reads this: please let me know if you think there are any problems with this, or some opportunity to make things simpler still with Webpack 2.

**Note:** If you're trying to read the diff, it makes much more sense if you ignore whitespace differences (because indentation changed). This URL will give an indentation-ignoring diff: https://github.com/aspnet/JavaScriptServices/pull/615/files?w=1


MarkPieszak:
LGTM! 👍 
Don't seeing anything off the bat that looks off.

liverpoolowen:
Brilliant! Any idea the size of the output files now it uses treeshaking?

SteveSandersonMS:
@LiverpoolOwen I wouldn't expect any significant size changes yet: you'd also need (at minimum) to be precompiling your views so that the set of API calls your code makes is known at compile time. I hope this is something we will get soon, but it depends on things like `ngc` compatibility with Angular Universal, Webpack, etc., and last time I checked it wasn't ready for that.

SteveSandersonMS:
This is now merged into `dev` and will be released as `generator-aspnetcore-spa` version 0.8.0 shortly.

PR #637, opened 2017-02-02
fix(ng-router): update to 3.4.5 for experimental version

MarkPieszak:
When everything is 4.x they'll all be on the same page / semantic versioning :)


SteveSandersonMS:
Thanks - merged.

I don't think we need to release a new version of `generator-aspnetcore-spa` to NPM for this though, because the version specifier was already `"^3.0.2"`, which means people creating new projects would already get 3.4.5 (or whatever is the latest 3.x version) anyway. Hope that's OK!

MarkPieszak:
You're totally right, I forget about that `^` sometimes :)
Sounds good!

Issue #669, opened 2017-02-13
Angular 2 SPA Docker: Can't access site


EduardTheThird:
Hi Guys

Hope you are well!

I've noticed strange behaviour regarding Docker containers and Angular2Spa where I can't access the site using the published IP port as usual. Opening the site on the published port, does not trigger activity on the Docker container. I did not see any obvious errors upon building the Docker image.

![docker1](https://cloud.githubusercontent.com/assets/6488837/22877120/96e15198-f1dc-11e6-942e-10508c89c727.PNG)
![docker2](https://cloud.githubusercontent.com/assets/6488837/22877124/9835328a-f1dc-11e6-9cc5-a223fd699496.PNG)

To ensure it is not something I changed on my project, I've created a freshly generated Angular2Spa (2.4.5) using generator-aspnetcore-spa 0.8.3 and generated a clean unmodified Docker image for Angular2Spa. 

The test docker image shown above is available from the Docker hub-> https://hub.docker.com/r/lightfleet/jssdocker/

Please let me know if you need more detail or if I can be of any assistance.

Thanks!

pysco68:
Hi there, 

I just had a similar issue (albeit with the Aurelia template). I solved it by adding an env variable:
ASPNETCORE_URLS=http://+:5000

Could you give that a try?

EduardTheThird:
Your suggestion saved our docker deployments, thank you @pysco68!

I can confirm that this worked for Angular 2. I've added a new pull request for the fix https://github.com/aspnet/JavaScriptServices/pull/687.

pysco68:
You're welcome! Glad I could help.
 Still debugging another issue in the same context atm ^^"

SteveSandersonMS:
Thanks for reporting this and submitting the PR.

As discussed in #687, we're removing the `Dockerfile` to be more consistent with other ASP.NET templates. Developers using Docker can add their own `Dockerfile` with their own config. As such, this is no longer applicable so I'll close it.

PR #687, opened 2017-02-16
Fix Docker for all templates.

EduardTheThird:
This fixes https://github.com/aspnet/JavaScriptServices/issues/669.

Thanks to @pysco68 for the tip!

### DIFF
--- a/templates/Angular2Spa/Dockerfile
+++ b/templates/Angular2Spa/Dockerfile
@@ -8,6 +8,8 @@ COPY . /app
 
 WORKDIR /app
 
+ENV ASPNETCORE_URLS http://+:5000
+
 RUN ["dotnet", "restore"]
 RUN ["dotnet", "build"]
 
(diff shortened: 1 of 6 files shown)


pysco68:
I think there's a better way to do so

EduardTheThird:
Hi @pysco68,

That is pure genius, thank you for the tip. I can confirm by adding  the suggested env variable that docker now works. What changes would you recommend?

pysco68:
Sorry, I was confusing something with the JS Service templates and my setup. 

No complaint at all! Great you put that together so quickly!

SteveSandersonMS:
Thanks for submitting this!

I know this might be not what you personally want to hear, but I'm afraid we're going to remove the `Dockerfile`s from these templates. This is because it's not standard to have them in the ASP.NET Core templates (the regular File->New ones don't), and having them in by default causes issues (e.g., #694). People who do want to use Docker can of course continue to add the `Dockerfile` like they would to any other ASP.NET project and it will still work. It should not affect you personally in your current project since you've already created your project.

As such, I'll mark this as closed. If you'd be interested in adding a page to the [list of recipes wiki](https://github.com/aspnet/JavaScriptServices/wiki) about enabling Docker, that would be great though!

Issue #705, opened 2017-02-23
TS2.2 doesn't like "@types/whatwg-fetch"


benaadams:
https://github.com/Microsoft/TypeScript/issues/4948#issuecomment-279109510 (Sounds like issue with isomorphic-fetch also)

The entry
```
"@types/whatwg-fetch": "0.0.28",
```
Needs to be removed from in fex. [ReactReduxSpa/package.json](https://github.com/aspnet/JavaScriptServices/blob/dev/templates/ReactReduxSpa/package.json#L14)

But then [domain-task](https://github.com/aspnet/JavaScriptServices/blob/dev/src/Microsoft.AspNetCore.SpaServices/npm/domain-task/package.json) doesn't like it as its a dependency 😞 

[Breaking-Changes#typescript-22](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#typescript-22)

benaadams:
Or maybe temporally change typescript from major version `"typescript": "^2.0.3",` to minor version `"typescript": "~2.0.3",`?

SteveSandersonMS:
Thanks for reporting this!

To fix this, I've updated the templates' TS dependencies to 2.2.1+. This meant that the `@types/whatwg-fetch` package references could be removed, since `fetch` is now included in the TS standard lib.

The updated templates will be published as version 0.8.6 shortly.

emirotin:
It would be nice if the updated typings have kept the parity with the `whatwg-fetch` types.
Specifically the typed version of `Body#json<T>()` would be nice to have: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/whatwg-fetch/index.d.ts#L39

kuncevic:
Similar story here https://github.com/angular/angular-cli/issues/5692 any thoughts?

PR #706, opened 2017-02-23
Typescript to minor match

benaadams:
Contributes to https://github.com/aspnet/JavaScriptServices/issues/705

Doesn't solve the compat problem; but does solve the issue with the templates at least for now.


benaadams:
May have done too many files

SteveSandersonMS:
Thanks for this. Rather than putting in this as a temporary fix, I've applied a more long term fix as described in #705. Hope that's OK!

Issue #667, opened 2017-02-12
React template: CSS Modules HMR not working


novascreen:
As soon as i add `?modules` to the `css-loader` I get the following message after a CSS change:
```
[HMR] The following modules couldn't be hot updated: (Full reload needed)
This is usually because the modules which have changed (and their parents) do not know how to hot reload themselves. See http://webpack.github.io/docs/hot-module-replacement-with-webpack.html for more details.
[HMR]  - ./~/css-loader?modules!./ClientApp/css/site.css
```
I couldn't find anything that i'm missing to make that work. As far as I understand this should work.
I also looked for an option to pass `reload=true` to `webpack-hot-middleware/client`, but nothing seems to be exposed for that.

SteveSandersonMS:
This has been reported a few times both to the `css-loader` and `webpack-dev-server` projects, e.g., https://github.com/webpack-contrib/css-loader/issues/186. It's an aspect of how `css-loader` is designed, rather than something controlled in this repo.

You're right that we don't currently expose a way to pass `reload=true` to the HMR client endpoint. That would be a good thing to add. Would you be interested in submitting a PR that extends `WebpackDevMiddlewareOptions` with some new options (e.g., `ClientOptions` of type `IDictionary<string, string>`), and correspondingly update `aspnet-webpack` so it adds any supplied option values to the client endpoint URL?

I'll close this as currently working as expected, but please feel free to post here further if you want to discuss submitting such a PR.

novascreen:
@SteveSandersonMS I'd be interested in giving this a try and think I've made out the spots to do the changes:
`src\Microsoft.AspNetCore.SpaServices\Webpack\WebpackDevMiddlewareOptions.cs`, `src\Microsoft.AspNetCore.SpaServices\npm\aspnet-webpack\src\WebpackDevMiddleware.ts` and add some docs to `src\Microsoft.AspNetCore.SpaServices\README.md`
I'm running into some trouble though when I try to `npm link` `aspnet-webpack` into one of the React templates. Firstly the templates seem to be missing some step to run a webpack build before the first `dotnet run`.  I fixed that by adding some scripts to `packages.json` to simplify this:
```
  "scripts": {
    "dotnet": "cross-env ASPNETCORE_ENVIRONMENT=Development dotnet run",
    "build": "webpack --config webpack.config.vendor.js && webpack"
  },
```
[...]

PR #714, opened 2017-02-25
Add option to configure Webpack Hot Middleware client

novascreen:
Addresses #667

I wasn't sure whether to create an interface for `HotModuleReplacementClientOptions` since we're basically just passing through options to a third party module and the options of that could change.
Please let me know if there is anything I can improve.


SteveSandersonMS:
Excellent work - thanks so much for this! It's now merged.

PR #721, opened 2017-02-28
Fixup MSBuild conversion

natemcmaster:
This is a partial fix for the solution not building with MSBuild.

~~This is blocked by https://github.com/Microsoft/msbuild/issues/1767~~


natemcmaster:
@SteveSandersonMS I remove the Web Sites projects because of that MSBuild issue. If you still want this to work in VS, we could have two solutions in the file, one that has the Web Site projects (for working in VS) and one without it (for CI builds).

SteveSandersonMS:
Thanks for this @natemcmaster - it looks great! It's also nice to have the NPM restoration phase implemented here rather than in a hidden SDK, because we might want to swap out `npm install` for `yarn` at some point to make the builds go through faster.

PR #722, opened 2017-02-28
Unify dependency versions across all projects

natemcmaster:
This puts all dependency versions into one file that his shared will all projects. This makes it easier to ensure versions are consistent.

@SteveSandersonMS I made this a separate PR from #721 as the change is not essential to the VS 2017 upgrade. We are moving towards this model in our other aspnet repos.


SteveSandersonMS:
Thanks @natemcmaster! This is a big improvement.

I did edit the commit a little when merging. I reverted the changes to the templates projects `.csproj` files (e.g, Angular2Spa, ReactSpa, etc.). That's because those aren't just regular .NET projects - they are the input to the system that builds Yeoman and `dotnet new` templates. As given, the tokenised values (e.g., `$(AspNetCoreVersion)` would have ended up in the template packages, which then would have generated projects that can't compile (because the `dependencies.props` file wouldn't be there). Also there's an argument that the dependency versions we use in the templates don't need to be the same as we use elsewhere, and we need to make separate decisions about when to change those dependency versions. So overall I think it's preferable to continue specifying those versions explicitly and independently. Hope that's OK!

natemcmaster:
Ah, thanks for clarifying. I wasn't 10% sure that the template projects were for. Thanks for cleaning it up 👍

PR #730, opened 2017-03-02
update tsconfig.js to use files instead of paths

### LINKED ISSUE #729: Latest version doesn't run in development mode
I just updated to the latest version 0.8.6 and when I run `yo aspnetcore-spa` to generate a new project using React I cannot run in debug mode. This seems to be an error with a `isomorphic-fetch` and `typescript`. This may be due to the upgrade to the new typescript version. Not sure if there's anything you can do or another package has to be updated. 

image

bmeverett:
Fixes #729. 

Changing `paths` to `files` in the `tsconfig.js` seems to fix the problem. I am no longer getting errors while in development mode.  This seems to be the correct way to do it based on the [documentation](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html). I see no reference to `paths` being used

### DIFF
--- a/templates/ReactReduxSpa/tsconfig.json
+++ b/templates/ReactReduxSpa/tsconfig.json
@@ -9,7 +9,7 @@
     "skipDefaultLibCheck": true,
     "lib": ["es6", "dom"],
     "types": [ "webpack-env" ],
-    "paths": {
+    "files": {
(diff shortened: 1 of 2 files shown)


bmeverett:
So I didn't have my environment variable set so the problem is still happening. I thought this might still be valid but I am also seeing some references to the `path` configuration option now.

PR #742, opened 2017-03-07
Duplicate identifiers in FetchData.tsx

### LINKED ISSUE #729: Latest version doesn't run in development mode

bmeverett:
I removed the `isomorphic-fetch` import from `FetchData.tsx`. This was causing duplicate types to be imported due to the upgraded typesecript version. Removing this gets rid of the errors and the FetchData page behaves the same. 

Fixes #729


bmeverett:
Not an issue. Was an error on my end as discussed in the issued. Closing.

Issue #739, opened 2017-03-06
Enable babel-plugin-transform-react-jsx-source in development for better warning messages


gaearon:
In [Create React App](https://github.com/facebookincubator/create-react-app), we enable [babel-plugin-transform-react-jsx-source](https://www.npmjs.com/package/babel-plugin-transform-react-jsx-source) in development configuration. I would suggest you to do the same.

It must **only** be enabled in development (since it produces bigger bundles). When it is enabled, React warnings get component stack traces with exact line numbers.

Example before:

![](https://pbs.twimg.com/media/CtCpNoiWcAAO9nk.jpg)

Example after:

![](https://pbs.twimg.com/media/CtCpPEMXgAAw1bq.jpg)

Hope you find this useful!

DanHarman:
I don't currently have time to make a pull request for this, but if anyone wants to add this feature either locally or contribute then this seems to work ok:

npm install --save-dev babel-plugin-transform-react-jsx-source

and modifying the webpack rule for babel-loader as below:

```js
module: {
            rules: [
                { test: /\.tsx?$/, include: /ClientApp/, use: { loader: 'babel-loader', options: { "plugins": isDevBuild ? ["transform-react-jsx-source"] : [] } } },
                { test: /\.tsx?$/, include: /ClientApp/, use: 'awesome-typescript-loader?silent=true' }
            ]
        },
```

DanHarman:
Addtionally wondering if we shouldn't be setting `cacheDirectory: true` in the babel options. Theoretically improves compilation times, but I don't know enough about this template to be sure it won't cause probs.

DanHarman:
ok decided to man up, pull request incoming :) The cache does speed up the build by 30-40%.

gaearon:
We use `cacheDirectory: true` in `babel-loader` options in CRA and it works great.

DanHarman:
That might have been where I got the idea from ;)

SteveSandersonMS:
Thanks for the suggestion. Since @DanHarman is implementing this in #749, I'll close this one.

BTW @gaearon thanks for the wonderful work on Redux. I especially like the docs - really easy and fun to read through.

SteveSandersonMS:
@gaearon Do you know if there's some way to make `babel-plugin-transform-react-jsx-source` work with source maps, or some other equivalent Webpack loader that supports compile-to-JS languages such as TypeScript?

[...]

PR #749, opened 2017-03-09
React Redux: Babel config enhancements

DanHarman:
Addressing issue #739 

- Added 'babel-plugin-transform-react-jsx-source' to development build
to add line numbers to debug output.
- Added 'cacheDirectory: true' to babel config which improves webpack
build time c. 40%.

### DIFF
--- a/templates/ReactReduxSpa/package.json
+++ b/templates/ReactReduxSpa/package.json
@@ -41,5 +41,8 @@
     "webpack": "^2.2.0",
     "webpack-hot-middleware": "^2.12.2",
     "webpack-merge": "^0.14.1"
+  },
+  "devDependencies": {
+    "babel-plugin-transform-react-jsx-source": "^6.22.0"
   }
 }
--- a/templates/ReactReduxSpa/webpack.config.js
+++ b/templates/ReactReduxSpa/webpack.config.js
@@ -17,7 +17,7 @@ module.exports = (env) => {
         },
         module: {
             rules: [
-                { test: /\.tsx?$/, include: /ClientApp/, use: { loader: 'babel-loader', options: { plugins: isDevBuild ? ["transform-react-jsx-source"] : [], cacheDirectory: true } } },
+                { test: /\.tsx?$/, include: /ClientApp/, use: { loader: 'babel-loader', options: { plugins: isDevBuild && ["transform-react-jsx-source"], cacheDirectory: true } } },
                 { test: /\.tsx?$/, include: /ClientApp/, use: 'awesome-typescript-loader?silent=true' }
             ]
         },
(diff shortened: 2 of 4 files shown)


DanHarman:
I guess it might be nice to apply this to the non-redux template too? If you would like me to do so, I can add that in too.

One question I have is how do you go about testing these changes in situ of the repo? I tested my mods in a project I created from the published template, but I'm not sure how I would go about doing this directly from the repo? (e.g. I can't run webpack in the repo as various files are missing in the template folder that webpack is expecting).

DanHarman:
hmm should maybe move the cacheDirectory setting into .babelrc rather than inlining with my other change (which needed to be in the webpack config to have convenient access to isDevBuild).

SteveSandersonMS:
Thanks for this!

> I guess it might be nice to apply this to the non-redux template too? If you would like me to do so, I can add that in too.

That would be great. I'll wait for an update then before merging this PR.

> One question I have is how do you go about testing these changes in situ of the repo? 

You can run the template projects directly. I don't know what missing files you encountered, but the following works:

```
cd templates/ReactReduxSpa
npm install
webpack --config webpack.config.vendor.js
webpack
dotnet restore
dotnet run
```

Optionally, also run `set ASPNETCORE_ENVIRONMENT=Development` (on Windows) or `export ASPNETCORE_ENVIRONMENT=Development` (on macOS/Linux) before `dotnet run` if you want HMR etc.

DanHarman:
Ok fixed up reactspa too.

One thing I noticed is that all the npm packages are dev dependencies in `packages.config`. Is that right for jquery? (maybe being able to import it rather than having to reference the js directly?)

SteveSandersonMS:
As far as I can tell, if we wire up `babel-plugin-transform-react-jsx-source` like this, it produces incorrect line numbers in the warnings. I guess the line numbers it uses refer to some intermediate `.jsx` representation, not the `.tsx` files.

As per [@gaearon's comment](https://github.com/aspnet/JavaScriptServices/issues/739#issuecomment-286228295), the line numbers come from Babel. I don't know if it's possible to get the original line numbers to flow through somehow (e.g., using the data from sourcemaps).

@DanHarman If you have any thoughts on how it can be made to work correctly, please let me know! If not, we'll probably have to not introduce this (since wrong line numbers would be worse than no line numbers).

SteveSandersonMS:
@DanHarman Since the consensus is that there isn't currently a way to get correct line numbers when the original source is `tsx` rather than `jsx`, I suggest we close this. Do you agree?

DanHarman:
Yes. It's a shame they aren't accurate as still useful as a clue, but not good enough to ship to people I agree.

One thing though, the cacheDirectory option I also added is worth keeping as it does significantly improve the webpack build time - unless you found issues with that? If you are happy with it I can create a separate PR with just that in it.

SteveSandersonMS:
> One thing though, the cacheDirectory option I also added is worth keeping as it does significantly improve the webpack build time

Yes, that would be great! I'll close this PR, but will be happy to merge a new one that just has the `cacheDirectory` bit in it. Thanks.

PR #760, opened 2017-03-14
Bubble up better error messages

TarogStar:
I was not able to track down the real problem without more detailed error messages, rather than just the stack.  This will give developers very useful information from webpack, which has very detailed messages.

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack/src/WebpackDevMiddleware.ts
+++ b/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack/src/WebpackDevMiddleware.ts
@@ -84,7 +84,7 @@ function attachWebpackDevMiddleware(app: any, webpackConfig: webpack.Configurati
             try {
                 aspNetWebpackReactModule = require('aspnet-webpack-react');
             } catch(ex) {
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Out of interest, could you provide an example of when this makes a difference?

In the cases I've tried, `exception.stack` already contains the information given by `exception.name` and `exception.message`. For example:

```
try {
    throw new Error('Some message');
} catch(ex) {
    console.log('Name: ' + ex.name);
    console.log('Message: ' + ex.message);
    console.log('Stack: ' + ex.stack);
}
```

... outputs:

```
Name: Error
Message: Some message
Stack: Error: Some message
    at (stack trace with line numbers, etc.)
```

So, the `name` and `message` are included in the `stack`. Are there cases where `stack` does not supply the `name`/`message` information?

SteveSandersonMS:
I'll close this due to lack of information (the conversation seems to have stopped). If anyone knows the reason why this change would be required, please let us know!

TarogStar:
@SteveSandersonMS Sorry I didn't get back to you.  I obviously didn't have access to the original exception, which was why I added this.  It was buried deep in some webpack created files and the exception was being swallowed by zone and I couldn't get it to stop automatically until it the exception was already lost unless I tried wading through seemingly hundreds of unrelated exceptions from the js es6 shims and the like.  All I got was the top message in my log.  With this change people would be able to see in their console exactly what is happening when they are using webpack.

Issue #768, opened 2017-03-16
Working example using Visual Studio 2017 and docker?


fbeltrao:
Hi there,

I am struggling trying to run the template application inside Docker.
Looks to me that JavaScriptServices can't find nodejs on the docker image microsoft/aspnetcore:1.1

I get the error "An exception of type 'System.AggregateException' occurred in System.Private.CoreLib.ni.dll but was not handled in user code" 

I tried to modify the Dockerfile adding nodejs the following way:
FROM microsoft/aspnetcore:1.1
RUN apt-get update
RUN apt-get -qq update
RUN apt-get install -y nodejs npm
#debian installs `node` as `nodejs`
RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
ARG source
WORKDIR /app
EXPOSE 80
COPY ${source:-obj/Docker/publish} .
ENTRYPOINT ["dotnet", "MyApp.dll"]

Still without success.

Using the following dockerfile works (because the docker image has nodejs installed):
FROM microsoft/aspnetcore-build
ARG source
WORKDIR /app
EXPOSE 80
COPY ${source:-obj/Docker/publish} .
ENTRYPOINT ["dotnet", "MyApp.dll"]

I just think it is overkill to have this 1.19GB image (compared to 268 MB from aspnetcore) just to run the aspnetcore + nodejs. 

Would it be possible to create a sample application where Docker is working?

mpriem:
+1

What a coincidence. I was troubleshooting the exact same issue. As Microsoft.AspNetCore.NodeServices.HostingModels.OutOfProcessNodeInstance.LaunchNodeProcess(ProcessStartInfo startInfo) refers to node by processname 'node', the Exception is thrown.

SteveSandersonMS:
The following worked for me:

```
FROM microsoft/dotnet:1.1-sdk-msbuild

RUN apt-get update
RUN wget -qO- https://deb.nodesource.com/setup_6.x | bash -
RUN apt-get install -y build-essential nodejs

COPY . /app

WORKDIR /app

RUN ["dotnet", "restore"]
RUN ["dotnet", "build"]

EXPOSE 5000/tcp

CMD ["dotnet", "run", "--server.urls", "http://0.0.0.0:5000"]
```

Note that it was also necessary to modify `Program.cs`, adding `.UseUrls("http://0.0.0.0:5000")` before `.Build()`.

PR #772, opened 2017-03-20
Added docker support to sample projects

fbeltrao:
Hi,

I've added Docker support to the sample projects. There is a problem with webpack in the React MusicStore sample.

This is related to issue #768.

The file DockerSupport.md explains the problem and the solution.

Please let me know if you need more information.

### DIFF
--- a/DockerSupport.md
+++ b/DockerSupport.md
@@ -1,0 +1,48 @@
+# Docker Support JavaScriptServices
+
+Using Visual Studio 2017 you can right click each project and Add Docker Support. Visual Studio will create a Dockerfile for the project and register it in the docker-compose section of your solution.
+The Dockerfile created by Visual Studio looks like this:
+```
+FROM microsoft/aspnetcore:1.1
+ARG source
+WORKDIR /app
+EXPOSE 80
+COPY ${source:-obj/Docker/publish} .
+ENTRYPOINT ["dotnet", "MusicStore.dll"]
+```
+
+The problem is that JavaScriptServices needs NodeJS to run, which is not available on the microsoft/aspnetcore:1.1 image. We need to install it before starting the application.
+Change the Dockerfile like this:
+```
+FROM microsoft/aspnetcore:1.1
+RUN apt-get update
+RUN apt-get install curl
+RUN curl -sL https://deb.nodesource.com/setup_6.x | bash
+RUN apt-get install -y build-essential nodejs
+ARG source
+WORKDIR /app
+EXPOSE 80
+COPY ${source:-obj/Docker/publish} .
+ENTRYPOINT ["dotnet", "MusicStore.dll"]
+```
+
+Docker support was added to all sample projects. The React MusicStore is not work properly due to a webpack error.
+
+## Running with Docker
+To run the application inside Docker, make sure that docker-compose is the Start-up project on your solution. Then, simply press F5 which will run all configured samples in a docker container.
+To browse go to the command line and type:
+```
+docker ps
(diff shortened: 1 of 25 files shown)


EduardTheThird:
Thank you for adding the docker samples! 

Will it be possible to add a example of using a Windows Container (microsoft/dotnet:nanoserver)?
I've been working on a Windows Container version. Currently it installs Node usings PowerShell and builds / publishes, however the site is not viewable once deployed in the container. 
It still needs to be optimized and moving to dotnet:nanoserver-runtime should help with the large size of this image.

```
FROM microsoft/dotnet:nanoserver

SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]

ENV NPM_CONFIG_LOGLEVEL info
ENV NODE_VERSION 6.10.0
ENV NODE_SHA256 39c809fe02863963a4bb1eaa0daec59abcc92c0331447c3d3e06392a7f72b79c

RUN Invoke-WebRequest $('https://nodejs.org/dist/v{0}/node-v{0}-win-x64.zip' -f $env:NODE_VERSION) -OutFile 'node.zip' -UseBasicParsing ; \
		if ((Get-FileHash node.zip -Algorithm sha256).Hash -ne $env:NODE_SHA256) {exit 1} ; \
    Expand-Archive node.zip -DestinationPath C:\ ; \
    Rename-Item -Path $('C:\node-v{0}-win-x64' -f $env:NODE_VERSION) -NewName 'C:\nodejs' ; \
    New-Item $($env:APPDATA + '\npm') ; \
    $env:PATH = 'C:\nodejs;{0}\npm;{1}' -f $env:APPDATA, $env:PATH ; \
    Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Control\Session Manager\Environment\' -Name Path -Value $env:PATH ; \
    Remove-Item -Path node.zip

COPY . /app

WORKDIR /app
RUN ["dotnet", "restore"]
RUN ["dotnet", "publish -o out"]

WORKDIR /app/out

ENV ASPNETCORE_URLS http://+:80
EXPOSE 80/tcp

ENTRYPOINT ["dotnet", "MusicStore.dll"]
CMD ["dotnet", "run", "--server.urls", "http://*:80"]
```

SteveSandersonMS:
@fbeltrao Thanks very much for this contribution!

I'm not sure if you're familiar with the history, but we used to include `Dockerfile`s, but specifically removed them for consistency with other ASP.NET templates, and to reduce the concept count - this was in commit c9526dd4d238898e98f923382a067a04f3f10ba6. I don't think we'll want to re-add that as a default feature in the templates.

That said, the information you've provided is really high quality and will be valuable to anyone who does want to use Docker with these kinds of projects. People often do ask about it, so it would be great to have an official place to maintain this information.

Would you be OK with this information being added to [the repo's wiki](https://github.com/aspnet/JavaScriptServices/wiki)? We're only just getting started there, but the idea is to build up a repository of knowledge around related topics, and this would be a perfect fit.

fbeltrao:
@SteveSandersonMS I was not aware of the Dockerfile consistency, sorry about that.
Of course you can add this information to the wiki page. Glad I can help.

Regards

SteveSandersonMS:
Thanks! I'll mark this as closed then.

Would you be willing to provide the content to go in this wiki page? https://github.com/aspnet/JavaScriptServices/wiki/DockerSupport

I think the topics people will want to know about would be:

 * What Dockerfile is needed for local development
 * What Dockerfile is needed for publishing (if different)
 * Any specific VS2017 instructions

Does that sound right?

PR #795, opened 2017-03-24
Update to latest released webpack 2

waynebrantley:
installing your currently react-redux-spa template and running does not work due to server side rendering.
I have webpack installed globally (which installs 2.2)
I read I should try to install aspnet-webpack globally, which failed because it requires 1.13 or rc.

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack/package.json
+++ b/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack/package.json
@@ -33,6 +33,6 @@
     "webpack": "^1.13.2"
   },
   "peerDependencies": {
-    "webpack": "^1.13.2 || ^2.1.0-beta"
+    "webpack": "^1.13.2 || ^2.2.0"
   }
 }


SteveSandersonMS:
Sorry, I don't understand this PR. I suspect it might not do what you're trying to make it do.

> I read I should try to install aspnet-webpack globally

Don't do that! Please instead tell whoever gave you that advice they are mistaken. In general, do *not* install NPM packages globally. Doing that just leads to "works-on-one-machine-but-not-another" scenarios, because you don't know what global packages you're relying on. The only time to install packages globally is if they provide a command-line tool that you want to make available globally. `aspnet-webpack` does not contain any command-line tools.

Regardless, the version spec you're proposing (`^1.13.2 || ^2.2.0`) is a subset of the existing one (`^1.13.2 || ^2.1.0-beta`). As in, anything that would be compatible with your new spec is already compatible with the existing one. So I don't see how this change could have the effect you want.

If I'm missing something important, please let me know! Otherwise can I close this?

SteveSandersonMS:
BTW sorry if my tone was a bit prickly there! I was trying to do a bunch of things at once, and didn't really slow down to express this well like I should have.

Issue #796, opened 2017-03-24
Support V8 Inspector Integration for Node.js


novascreen:
It would be great if we could debug the Node instance with Chrome dev tools:

https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js

Or maybe make it possible to pass any arguments to the Node process?

SteveSandersonMS:
Agreed. Currently we support `node-inspector` (https://github.com/aspnet/JavaScriptServices/blob/dev/src/Microsoft.AspNetCore.SpaServices/README.md#debugging-your-projects), but it would be better if we changed this to pass `--inspect` instead of `--debug` to support newer versions of Node.

Do you have any interest in submitting a pull request that changes how `OutOfProcessNodeInstance.cs` deals with the `launchWithDebugging` flag? It should be changed to pass `--inspect` to the Node process instead of `--debug`.

> Or maybe make it possible to pass any arguments to the Node process?

Agreed that would also be good, but since we already have a defined API for starting in debugging mode, we should consider "arbitrary arguments" to be a separate feature.

novascreen:
Sounds good. I can look into submitting a pull request. Do you think it makes sense to add another option for break on first line?

SteveSandersonMS:
> Do you think it makes sense to add another option for break on first line?

Since (as a separate work item) we'll add the ability to pass arbitrary params, we're probably OK without additional startup options. Then at least if Node changes its debugging option names again, we won't have another issue to fix :)

SteveSandersonMS:
> Sounds good. I can look into submitting a pull request

Excellent - thanks! I'll close this issue then, and will await your PR.

novascreen:
> Since (as a separate work item) we'll add the ability to pass arbitrary params, we're probably OK without additional startup options
I was thinking the same thing. I've already created that PR: #804 
Btw. I have another PR waiting: #714

PR #804, opened 2017-03-25
Support V8 Inspector Integration for Node.js

novascreen:
Changes deprecated `--debug` to `--inspect` when using `LaunchWithDebugging`.

I have not implemented `inspect-brk`. Node 6 uses `debug-brk` so maybe this is enough for this feature and details like that might be better addressed by a separate feature which allows users to pass any parameter, as discussed in #796.

Node logs 4 separate messages for the debugger initially, so I wasn't sure what you would do to handle those. I've decided to catch any logs related to the debugger and log them as warnings.

One of the messages is a warning, but terminated the dotnet process, so I created a separate check for that (`StartsWith("Warning"`) and log that as a warning as well.

Let me know if there is anything you'd like me to improve.

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/HostingModels/OutOfProcessNodeInstance.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/HostingModels/OutOfProcessNodeInstance.cs
@@ -27,16 +27,6 @@ namespace Microsoft.AspNetCore.NodeServices.HostingModels
         protected readonly ILogger OutputLogger;
 
         private const string ConnectionEstablishedMessage = "[Microsoft.AspNetCore.NodeServices:Listening]";
-        private const string DebuggingStartedMessageFormat = @"-----
-*** Node.js debugging is enabled ***
-{0}
-
-To debug, run:
-   node-inspector{1}
-
-If you haven't yet installed node-inspector, you can do so as follows:
-   npm install -g node-inspector
------";
         private readonly TaskCompletionSource<object> _connectionIsReadySource = new TaskCompletionSource<object>();
         private bool _disposed;
         private readonly StringAsTempFile _entryPointScript;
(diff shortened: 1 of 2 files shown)


novascreen:
I just noticed that a `DeprecationWarning` thrown from Node also terminates the dotnet process, i could add that to the check and fix that as part of the PR or create a new one in which I take care of both warnings. Or maybe this should be handled somewhere else?

SteveSandersonMS:
Thanks - this is great too :) Merged.

I did make one small tweak to remove the `IsWarning` check because it would have been a breaking change (it would mean that lines starting `Warning:` no longer get passed to `OnErrorDataReceived`, which might have been overridden to do something with that data). Although we can make breaking changes if they are important, I think this change wasn't strictly required, so it's better to go ahead without it.

Everything to do with how you got `--inspect` running, and updated the docs, was all perfect - thanks so much!

novascreen:
@SteveSandersonMS I see this was recently published. But i don't understand how i would catch that warning now so the dotnet process doesn't fail instantly when it comes across that warning?
Maybe should specifically add `Warning: This is an experimental fe
ature and could change at any time.` to `IsDebuggerMessage`?

SteveSandersonMS:
@novascreen Yes, I'm seeing that now too. It didn't happen when I originally tried this back in March, but it does now on more up-to-date versions of Node. It's unfortunate that we don't have a way of knowing exactly which warning messages relate  to the debugger, given that future versions of Node might rephrase this in arbitrary ways.

So yes, we do need to capture that warning for this to work. I'd rather the code be more specific than just matching everything that starts with `Warning:`, because we don't want to suppress unrelated warnings.

Are you open to submitting a follow-up PR that catches the specific warning text we're seeing now?

novascreen:
for sure, I'll do that

PR #833, opened 2017-03-31
Console type needs exe

benaadams:
`Project Sdk="Microsoft.NET.Sdk"` needs `<OutputType>exe</OutputType>` or it thinks its a library


SteveSandersonMS:
Thanks @benaadams!

PR #853, opened 2017-04-09
Update Aurelia SPA template

MeirionHughes:
* Updates to new aurelia-webpack-plugin (Webpack 2). 
* Enables webpack hmr support by default

*todo:*

* [x] needs review (@EisenbergEffect or @niieani)
* [x] check production build

### DIFF
--- a/templates/AureliaSpa/webpack.config.js
+++ b/templates/AureliaSpa/webpack.config.js
@@ -1,45 +1,51 @@
-var isDevBuild = process.argv.indexOf('--env.prod') < 0;
-var path = require('path');
-var webpack = require('webpack');
-var AureliaWebpackPlugin = require('aurelia-webpack-plugin');
+const path = require('path');
+const webpack = require('webpack');
+const { AureliaPlugin } = require('aurelia-webpack-plugin');
 
-var bundleOutputDir = './wwwroot/dist';
-module.exports = {
-    resolve: { extensions: [ '.js', '.ts' ] },
-    entry: { 'app': 'aurelia-bootstrapper-webpack' }, // Note: The aurelia-webpack-plugin will add your app's modules to this bundle automatically
-    output: {
-        path: path.resolve(bundleOutputDir),
-        publicPath: '/dist',
-        filename: '[name].js'
-    },
-    module: {
-        loaders: [
-            { test: /\.ts$/, include: /ClientApp/, loader: 'ts-loader', query: { silent: true } },
-            { test: /\.html$/, loader: 'html-loader' },
-            { test: /\.css$/, loaders: [ 'style-loader', isDevBuild ? 'css-loader' : 'css-loader?minimize' ] },
-            { test: /\.(png|woff|woff2|eot|ttf|svg)$/, loader: 'url-loader?limit=100000' },
-            { test: /\.json$/, loader: 'json-loader' }
+module.exports = ({ prod } = {}) => {
+    const isDevBuild = !prod;
+    const isProdBuild = prod;
+    const bundleOutputDir = './wwwroot/dist';
+
+    return {
+        resolve: {
+            extensions: [".ts", ".js"],
+            modules: ["ClientApp", "node_modules"],
+        },
+        entry: { 'app': 'aurelia-bootstrapper' },
+        output: {
+            path: path.resolve(bundleOutputDir),
+            publicPath: "/dist/",
+            filename: '[name].js'
+        },
+        module: {
+            rules: [
+                { test: /\.css$/i, use: [isDevBuild ? 'css-loader' : 'css-loader?minimize'] },
+                { test: /\.html$/i, use: ["html-loader"] },
+                { test: /\.ts$/i, loaders: ['ts-loader'], exclude: path.resolve(__dirname, 'node_modules') },
+                { test: /\.json$/i, loader: 'json-loader', exclude: path.resolve(__dirname, 'node_modules') },
+                { test: /\.(png|woff|woff2|eot|ttf|svg)$/, loader: 'url-loader', query: { limit: 8192 } }
+            ]
+        },
+        plugins: [
+            new webpack.DefinePlugin({ IS_DEV_BUILD: JSON.stringify(isDevBuild) }),
+            new webpack.DllReferencePlugin({
+                context: __dirname,
+                manifest: require('./wwwroot/dist/vendor-manifest.json')
+            }),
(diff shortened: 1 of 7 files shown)


templates/AureliaSpa/webpack.config.js
```diff
@@ -1,24 +1,28 @@
...
         filename: '[name].js'
     },
-    module: {
-        loaders: [
-            { test: /\.ts$/, include: /ClientApp/, loader: 'ts-loader', query: { silent: true } },
-            { test: /\.html$/, loader: 'html-loader' },
-            { test: /\.css$/, loaders: [ 'style-loader', 'css-loader' ] },
-            { test: /\.(png|woff|woff2|eot|ttf|svg)$/, loader: 'url-loader?limit=100000' },
-            { test: /\.json$/, loader: 'json-loader' }
+    module: {                                   // (4)
```
niieani: `// (4)` ? ;)

templates/AureliaSpa/webpack.config.js
```diff
@@ -1,24 +1,28 @@
-var isDevBuild = process.argv.indexOf('--env.prod') < 0;
-var path = require('path');
-var webpack = require('webpack');
-var AureliaWebpackPlugin = require('aurelia-webpack-plugin');
+const isDevBuild = process.argv.indexOf('--env.prod') < 0;
```
niieani: Seems quite ugly. The `--env.YYY` option is there so you can `module.exports = function(env)` where `env` will be an object containing the passed in `env.YYY`.

templates/AureliaSpa/webpack.config.js
MeirionHughes: removed

templates/AureliaSpa/webpack.config.js
MeirionHughes: switched to what the others are using: 
```
module.exports = (env) => {
    const isDevBuild = !(env && env.prod);
```

EisenbergEffect:
Deferring to @niieani When he says it's good, then I'm happy :)

templates/AureliaSpa/webpack.config.js
```diff
@@ -1,45 +1,52 @@
...
-        filename: '[name].js'
-    },
-    module: {
-        loaders: [
-            { test: /\.ts$/, include: /ClientApp/, loader: 'ts-loader', query: { silent: true } },
-            { test: /\.html$/, loader: 'html-loader' },
-            { test: /\.css$/, loaders: [ 'style-loader', 'css-loader' ] },
-            { test: /\.(png|woff|woff2|eot|ttf|svg)$/, loader: 'url-loader?limit=100000' },
-            { test: /\.json$/, loader: 'json-loader' }
+module.exports = (env) => {
```
niieani: You can set a default empty object `(env = {}) =>` here and you won't have to `env && ` every time it's used :) Also, it might be arguably better to use destructuring here (maybe even with reassignment), like: `({prod} = {})`.

templates/AureliaSpa/webpack.config.js
```diff
@@ -1,45 +1,52 @@
...
+    const isDevBuild = !(env && env.prod);
+    const isProductionBuild = !isDevBuild;
+    
+    const bundleOutputDir = './wwwroot/dist';
+
+    return [{
```
niieani: Why return an array?

templates/AureliaSpa/webpack.config.js
MeirionHughes: done. :)

templates/AureliaSpa/webpack.config.js
MeirionHughes: remnant of either the old config or from the one of the other templates. /removed.

MeirionHughes:
@niieani I've checked the production build with:

```shell
webpack --config webpack.config. vendor.js -p
webpack -p
dotnet run --env Production
```

it appears to be working fine (named slots working too).

nmocruz:
Why not use `includeAll: "ClientApp" ` in AureliaPlugin instead of `PLATFORM.moduleName` in every module?

MeirionHughes:
includeAll was added to make it easier to migrate older projects; moving forward though, it places some limitations on how your code can be optimised and so you should generally use PLATFORM.moduleName.

jods4:
@nmocruz 
I created `includeAll` as an easy way to migrate your existing project. Basically it means: just include all files in my source folder and be done.

It does have drawbacks, though:
- You'll have a larger output because it prevents quite a few optimizations on all your files, such as: unused code removal, module renaming, exports renaming, tree shaking.
- It prevents code splitting.
- It does not validate at build time that your module strings actually exist.

Overall you're missing out quite a lot on what webpack can offer. This is like going back to an AMD + gulp build that is just taking all files in all source and libs and concat them together in a single big file.

nmocruz:
@jods4 Ok, make sense. I need to get used to the idea of use PLATFORM.moduleName('my-module'). 
I was really unhappy with this change, after behing using just a string for some time. Is not so bad but shorter could be better.

jods4:
For libraires it's important to have a common, easily recognizable syntax (because JS is hard to analyze).
In your own code it might be different and in fact I already have everything required in place for you to define shorter synonyms.

You should also find that the places where module names appear are usually quite few.

nmocruz:
> You should also find that the places where module names appear are usually quite few.
yes, looks a good idea. I understand that PLATFORM.moduleName is a much more clear way but after the old approach it looks like repeating PLATFORM.moduleName just to be webpack friendly. 
Anyway, congratulation for this.

EisenbergEffect:
@SteveSanderson This is approved and ready to go on our side.

ClaudioNunes:
@SteveSanderson any estimate on when this merge will occur?

ry8806:
I think this account is probably the one he uses (its more active) @SteveSandersonMS 
Looking forward to seeing this merged soon!!!

ClaudioNunes:
@SteveSandersonMS  @EisenbergEffect  Any updates?

EisenbergEffect:
@SteveSandersonMS We definitely need this merged ASAP...please.

SteveSandersonMS:
@MeirionHughes Great job on this! A really clean, straightforward PR. Also I'm impressed by the improvements to Aurelia itself - the new HMR functionality works brilliantly.

I'll now work on publishing the updated template package to NuGet.

EisenbergEffect:
Thanks @SteveSandersonMS !

PR #894, opened 2017-04-25
V1.0.x

trongphan:
Browser to remember client side routing history and also server side routing.

Assume we have client side routing as below:
http://localhost:5000/
http://localhost:5000/counter
http://localhost:5000/fetchdata

We also have the routing as below which is not part of client side routing. This will make a request to server to Home Controller and TP action
http://localhost:5000/home/tp

Every work fine except the browser does not keep history of this server side routing when we access any client side routing park.

For instance, I access http://localhost:5000/home/tp
then http://localhost:5000/
then http://localhost:5000/counter

When I clieck Back in browser, it stop at  http://localhost:5000/. It does not track http://localhost:5000/home/tp in history.

I guess React-Router have remove all history from browser to manage history in their code.

Thanks,

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices/Content/Node/prerenderer.js
+++ b/src/Microsoft.AspNetCore.SpaServices/Content/Node/prerenderer.js
@@ -71,7 +71,8 @@
 	// a certain flag is attached to the function instance.
 	function renderToStringImpl(callback, applicationBasePath, bootModule, absoluteRequestUrl, requestPathAndQuery, customDataParameter, overrideTimeoutMilliseconds) {
 	    try {
-	        var renderToStringFunc = findRenderToStringFunc(applicationBasePath, bootModule);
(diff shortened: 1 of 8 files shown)


SteveSandersonMS:
Thanks @trongphan for getting in touch!

Would I be right in thinking that this is a bug/support report rather than a pull request? The code in the PR appears to be commits that are already in `dev`.

> When I clieck Back in browser, it stop at http://localhost:5000/. It does not track http://localhost:5000/home/tp in history.

I understand what you're reporting, but am not able to reproduce that issue. With the current React templates, it is possible to click 'back' and navigate back beyond your first entry point on the site as expected. Perhaps this was a bug in the previous React router - the current templates recently moved to React Router v4 which was pretty much a total rewrite of the router, so there's a good chance that edge-case bugs are fixed even if by accident.

Since this appears to be OK, I'll mark this as closed. But if you can still reproduce the problem with the updated templates, could you please file a new bug report?

PR #921, opened 2017-05-04
Templates: update React's component's State and Property type

stephtr:
The React templates used `any` and `void` as State and Property types for React.Component, even though it should be `{}`.


SteveSandersonMS:
Thanks @stephtr - this looks good!

antmdvs:
@stephtr Why was `void` considered incorrect? Just curious. Thanks!

stephtr:
That question was best answered in [TypeScript #15409](https://github.com/Microsoft/TypeScript/issues/15409): `props` is always an object, even without properties.

Issue #919, opened 2017-05-04
The Vue template should be using Official Companion Libraries


ferrx:
I saw that `av-ts` is being used in the Vue template. See the template's package json here: https://github.com/aspnet/JavaScriptServices/blob/dev/templates/VueSpa/package.json

[This Vue guide](https://vuejs.org/v2/guide/typescript.html#Class-Style-Vue-Components) mentions [vue-class-component](https://github.com/vuejs/vue-class-component) which is officially supported through [contributions by the Vue author](https://github.com/vuejs/vue-class-component/graphs/contributors).

Any thoughts on switching?

SteveSandersonMS:
I think we chose to use `av-ts` because Vue natively has pretty poor TypeScript support, even with `vue-class-component`.

However, if you think the template can be improved, you may well be right! If you have a pull request that improves the template while also not adding complexity, we'd be very happy to take a look.

ferrx:
@SteveSandersonMS Thanks, I will give it a shot. Are the directions for contributing a template up to date? From the readme.md:

> Cloning the repo
> Running dotnet restore at the repo root dir
> Going to whatever sample or template you want to run (for example, cd templates/Angular2Spa)
> Restoring NPM dependencies (run npm install)
> Launching it (dotnet run)

The wwwroot directory never gets populated. Should there be a webpack step?

ferrx:
Found this in the yo build src:
```
        this.installDependencies({
            npm: true,
            bower: false,
            callback: () => {
                this.spawnCommandSync('dotnet', ['restore']);
                this.spawnCommandSync('./node_modules/.bin/webpack', ['--config', 'webpack.config.vendor.js']);
                this.spawnCommandSync('./node_modules/.bin/webpack');
            }
        });
```
After performing `webpack --config webpack.config.vendor.js` and `webpack` I was able to `dotnet run` and view the site.

PR #923, opened 2017-05-05
Switched from av-ts to vue-property-decorator and vue-class-component

ferrx:
For more on these packages, there is a good guide here: https://alligator.io/vuejs/typescript-class-components/

[vue-class-component](https://github.com/vuejs/vue-class-component) is an `official companion library` to VueJS ([see here](https://vuejs.org/v2/guide/typescript.html#Class-Style-Vue-Components)) with contributor support from the VueJS author.

This PR simply makes the replacement. Ref #919


SteveSandersonMS:
Thanks - this is a great improvement! It's now merged.

PR #931, opened 2017-05-12
Add core-js dependency

astegmaier:
When I run webpack on the angular4 branch I get the error:
```
Module not found: Error: Can't resolve 'core-js/shim' in '/Users/Andrew/Projects/angular4-test/ClientApp'
     @ ./ClientApp/boot-server.ts 3:0-23
```
Others have reported the same problem (see #930). It appears that adding core-js to project.json fixes this problem--and that's all this PR does.

### DIFF
--- a/templates/Angular2Spa/package.json
+++ b/templates/Angular2Spa/package.json
@@ -22,6 +22,7 @@
     "aspnet-webpack": "^1.0.17",
     "awesome-typescript-loader": "^3.0.0",
     "bootstrap": "^3.3.7",
+    "core-js": "^2.4.1",
     "css": "^2.2.1",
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks for posting this! It turns out the issue was that `boot-server.ts` was assuming `core-js` would be installed, but it was only being installed if you enabled unit tests, because it was `karma` bringing in that dependency.

However, there shouldn't have been a runtime dependency on the full `core-js` package anyway, because it's sufficient just to use `reflect-metadata` like we do in `boot-client.ts`. So rather than adding `core-js` to `package.json`, I'm just going to remove the `core-js` reference from `boot-server.ts`. Hope that's OK!

PR #937, opened 2017-05-15
Added nonce to prerender script output to support CSP

nilsgs:
Currently we need to use `script-src 'unsafe-inline'` in the Content Security Policy header to get access to the global js object added to the page by the prerenderer.

I've added an optional `asp-prerender-nonce` attribute, to overcome this limitation.


buvinghausen:
This has been sitting here for 5 months I want this functionality as allowing unsafe-inline in your CSP is pretty much as good as not having a CSP....

SteveSandersonMS:
Since we're moving away from the use of `Globals` in prerendering (because the newer middleware-based approach used in the Angular CLI and create-react-app templates generates the *entire* HTML doc, not just some elements that are injected into a larger doc), there probably isn't a strong need for this any more. Developers get to control their entire prerendered page and can therefore do whatever they want to insert CSP attributes, etc. I'll therefore close this - hope that's OK, and thanks again for the original contribution!

PR #941, opened 2017-05-16
Outputs vs-2017.3.host.json file

madskristensen:
This should now write out the correct file for VS to pick up. The icon is missing since I'm not totally sure about how we should inject that into the _.template.config_ folder. Also, is there a description of each template we can use?


SteveSandersonMS:
Thanks! This is now merged, along with the icons, and a vague attempt at a `description` property.

Later on I'll change the sources so that instead of generating `.template.config` contents at build time, they are just plain files in source control. Then it will be easier to make per-template tweaks. I'll just change the Yeoman template generator to exclude the `.template.config` dir entirely.

SteveSandersonMS:
Also I'll push an updated template package later today, hopefully.

PR #951, opened 2017-05-18
Added IE polyfills to support IE9-11 (fix apply on undefined error)

AmrineA:
Added IE polyfills for IE9-11

- new ClientApp/polyfills/polyfills.ts file containing polyfills
- added core-js to package.json
- added core-js to webpack.config.vendor.js
- added polyfills.ts import to ClientApp/boot-client.ts


AmrineA:
Will resubmit with additional updates

Issue #926, opened 2017-05-10
DI Error in IE following shim installation


im1dermike:
I had created issue #891 where I was getting a _'WeakMap' is undefined_ error when I tried to load the ASP.NET Core Angular 2 app in IE10. @SteveSandersonMS suggested I import `core-js` which resolved the WeakMap issue, but now I'm getting this error:

> SCRIPT5022: Error: DI Error
> vendor.js, line 104835 character 13

The app works fine in all other browsers.

MarkPieszak:
You need to have the correct shims to for IE support lower than IE11.
See here for a reference of what you'd need: https://github.com/MarkPieszak/aspnetcore-angular2-universal/blob/master/Client/polyfills/polyfills.ts#L6-L24

im1dermike:
@MarkPieszak I ran the following:

    npm install --save classlist.js
    npm install --save web-animations-js

Then changed the top of my boot-client.ts file to look like this:

    import 'core-js/shim';
    import 'core-js/es6/symbol';
    import 'core-js/es6/object';
    import 'core-js/es6/function';
    import 'core-js/es6/parse-int';
    import 'core-js/es6/parse-float';
    import 'core-js/es6/number';
    import 'core-js/es6/math';
    import 'core-js/es6/string';
    import 'core-js/es6/date';
    import 'core-js/es6/array';
    import 'core-js/es6/regexp';
    import 'core-js/es6/map';
    import 'core-js/es6/set';
    import 'classlist.js';  // Run `npm install --save classlist.js`.
    import 'web-animations-js';  // Run `npm install --save web-animations-js`.
   
    import 'angular2-universal-polyfills/browser';
    import { enableProdMode } from '@angular/core';
    import { platformUniversalDynamic } from 'angular2-universal';
    import { AppModule } from './app/app.module';
    import 'bootstrap';
    const rootElemTagName = 'app'; // Update this if you change your root component selector

The same issue remains.

im1dermike:
Any update on this?  It's not really OK that this library doesn't work in IE10.

SteveSandersonMS:
We have an existing issue tracking support for older IE versions at #484. I'll close this as a duplicate of that.

> It's not really OK that this library doesn't work in IE10.

The template isn't meant to support all legacy browsers by default, as that would bloat the resulting application and not everyone needs that support. However I recognise that we could use clear docs on what polyfills are needed for people who do want to support legacy browsers.

PR #952, opened 2017-05-18
Update boot-client.ts to handle IE 11 shim issues

ADefWebserver:
In IE 11, when you get error like this:
Unable to get property 'apply' of undefined or null reference

In functions like this:
```
function combine(options) {
    return ((Object)).assign.apply(((Object)), [{}].concat(options));
}	
```

in the vendor.js file

This article tells you to use core-js because Angular already has a dependency on it:
http://stackoverflow.com/questions/41276692/angular2-ie11-unable-to-get-property-apply-of-undefined-or-null-reference

You add:
import "core-js/client/shim";

to the top of boot-client.ts

and the problem goes away. 

This issue may be related:
https://github.com/aspnet/JavaScriptServices/issues/926

### DIFF
--- a/templates/AngularSpa/ClientApp/boot-client.ts
+++ b/templates/AngularSpa/ClientApp/boot-client.ts
@@ -1,3 +1,4 @@
+import "core-js/client/shim";
 import 'reflect-metadata';
 import 'zone.js';
 import { enableProdMode } from '@angular/core';


AmrineA:
I can confirm that this change does resolve the error referenced.

im1dermike:
This did not resolve my issue #926

SteveSandersonMS:
Thanks for contributing this. Since the Angular template has now moved to being Angular CLI-based, we no longer have code in this repo to determine what polyfills are placed there - we take the output from Angular CLI. As such this no longer applicable, but thanks very much for getting involved!

PR #958, opened 2017-05-19
API route mapping to get 404 instead of Home/Index

AmrineA:
Changed the route mapping so that calls to API methods that don't exist return 404 and not a page


SteveSandersonMS:
Thanks for the suggestion! This is a bit too specific for us to include by default. However I'm glad you've got a solution that meets your own needs, and if anyone else raised this requirement, I'll point them to your PR as a suggestion solution. Hope that's OK!

PR #966, opened 2017-05-21
Organize webpack config to match as much as possible

manigandham:
Removed extra variables and organized both webpack config files to match as much as possible to make it easier to learn and compare differences.

### DIFF
--- a/templates/VueSpa/webpack.config.js
+++ b/templates/VueSpa/webpack.config.js
@@ -1,13 +1,13 @@
 const path = require('path');
 const webpack = require('webpack');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const extractCSS = new ExtractTextPlugin('site.css');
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks @manigandham for submitting this, and apologies for the slow response!

I think that sometime between when you submitted this and now, we must have refactored the Webpack configs already, because looking at the patterns across them all now, the proposed changes in this PR would no longer make the Vue template config consistent with the others. It seems that these changes would now make it *less* consistent with others, even if they are perfectly good changes on their own. The one thing that would still be applicable is removing the duplicated `bundleOutputDir` variable from the Vue template's `webpack.config.js`, which I'll do separately.

So again I apologise that my slow response has made this PR less applicable that it would originally have been. I'll mark it as closed, since things are now more consistent as they are anyway. Hope that's OK, and please continue to submit other PRs if you wish to contribute! Thanks.

manigandham:
No worries, thanks for the update.

PR #974, opened 2017-05-24
Fix Node debugging by catching warning

novascreen:
Following up on #804 

Fixes the debugger which causes the .NET Core app to crash due to a Node warning. 

Unfortunately it's hard to make this really future proof, because the wording and structure of the warnings regarding debugging could easily change in future versions. I still think it would be nice to be able to treat all warnings from Node differently from errors, but no idea how to approach that. Cause for a separate discussion I suppose.


SteveSandersonMS:
Thanks @novascreen!

PR #982, opened 2017-05-26
Re-add bootstrap import

danobri:
In the update to Angular 4 ([PR #800](https://github.com/aspnet/JavaScriptServices/issues/800)), the bootstrap import got removed from boot.client.ts, causing the navbar-toggle menu to no longer work in the template.  Re-adding so navbar-toggle is functional again.


SteveSandersonMS:
Thanks! This is now merged.

Issue #988, opened 2017-05-29
Improve error page while server-side rendering.


rosieks:
Error pages in ASP.NET are awesome - they provide nice looking information about error that happened on the server. It would be nice if we could have the same with JavaScriptServices. Right now I see two places where this could be improved.
* It would be nice if the JavaScript stack trace could show references to TS files instead of bundled boot-server.js
* Authors of React assumed that font used to render error will be fixed-width. Also it would be nice to have that in colors like it is when you run it in terminal.

![image](https://cloud.githubusercontent.com/assets/1107959/26557283/4f85a5f8-44a0-11e7-8d56-cdffeb9cb487.png)

SteveSandersonMS:
Thanks for the suggestion. I'll close this in favour of #1000.

PR #1000, opened 2017-05-31
Unified way to report node invocation errors

rosieks:
There is the different way how node invocation errors are reported when using `HttpNodeInstance` and `SocketNodeInstance`. This PR unify the way how such errors are reported and lets create middleware that to handle `NodeInvocationException` and display dedicated error page for node errors #988


SteveSandersonMS:
Thanks - this looks good!

PR #1003, opened 2017-06-01
Added source-map support for server-side code

rosieks:
This is another piece that improves developer experience mentioned in #988. This PR adds source-map support for JavaScript that run on server-side.

### DIFF
--- a/templates/ReactReduxSpa/ClientApp/boot-server.tsx
+++ b/templates/ReactReduxSpa/ClientApp/boot-server.tsx
@@ -1,3 +1,4 @@
+import 'source-map-support/register';
 import * as React from 'react';
 import { Provider } from 'react-redux';
 import { renderToString } from 'react-dom/server';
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
Thanks for the suggestion!

After reading up on https://www.npmjs.com/package/source-map-support, I'm not sure this is an additional dependency that we'd want to enable by default. I appreciate its benefits, but the fact that it changes some default Node behaviours (e.g., around uncaught exceptions) makes it risky. It's the sort of thing that developers can easily add to their own projects if they need it, and then at least they know what they've done and can take responsibility for now having certain nonstandard behaviours. I'm aware of the `handleUncaughtExceptions` and `hookRequire` flags but it's yet more concepts for people to make sense of so not ideal to put in by default.

Are you OK if we close this?

rosieks:
Yes, that sounds reasonable.

PR #1012, opened 2017-06-05
Moved FormsModule and HttpModule to app.modules.shared

AmrineA:
Moved FormsModule and HttpModule from from app.module.client to app.module.shared


SteveSandersonMS:
Thanks for contributing this!

Based on @MarkPieszak's advice, I extended your PR a bit to make the shared module into a real `@NgModule`. This is now committed in f3b7103c83ef38d787c2e66bc761a1c4c562937a.

Issue #509, opened 2016-12-12
Add option to React/Redux template to include Jest testing


LovedByTheLord:
Hi, 

I saw that an option to add client side unit testing was added to the Angular template.  I would like to suggest a similar option for adding Jest to the React/Redux template.

Thanks!

SteveSandersonMS:
If you can implement this in a really simple and clean way, we could definitely consider that! Pull requests welcome.

PR #1041, opened 2017-06-17
Added sample "jest" tests for counter reducer #509

rosieks:
This PR fills the gap between Angular and React-Redux template by adding test library and sample tests to the second one. #509

### DIFF
--- a/templates/ReactReduxSpa/ClientApp/store/Counter.spec.ts
+++ b/templates/ReactReduxSpa/ClientApp/store/Counter.spec.ts
@@ -1,0 +1,22 @@
+﻿import 'jest';
+import { reducer } from './Counter';
+
+describe('Coutner reducer', () => {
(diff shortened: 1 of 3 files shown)


SteveSandersonMS:
Thanks for the suggestion!

These `jest` tests are not really equivalent to the `karma` tests in the Angular project, because the `karma` ones are testing at a UI level by automating the browser, whereas these `jest` tests are purely unit tests that don't describe the UI at all.

If we were to add sample tests to the Redux project, I think we'd want to add higher-level UI tests for equivalence. But that would still only be if those tests and its setup/config were all simple enough to be easily understood by a total newcomer.

Is that OK? Do you mind if I close this PR?

rosieks:
That make sense.

SteveSandersonMS:
Thanks!

Issue #897, opened 2017-04-25
How about add "private": true to package.json?


g0t4:
How about add `"private": true` to `package.json`? 

This will avoid the following warnings on npm install:

```bash
npm WARN Skills_Web@0.0.0 No description
npm WARN Skills_Web@0.0.0 No repository field.
npm WARN Skills_Web@0.0.0 No license field.
```

and of course avoid publishing, as web projects tend to be private.

SteveSandersonMS:
Great idea - would you be willing to submit a pull request? Thanks!

g0t4:
Yup

PR #1044, opened 2017-06-18
Added "private": true to package.json

rosieks:
This PR adds `"private": true` to _package.json_ in all templates like it was discussed in #897


SteveSandersonMS:
Thanks @rosieks!

Issue #1066, opened 2017-06-26
New React/Redux project: webpack fails with Typescript errors


jimsmart:
Hi,

New project freshly generated with 'React with Redux' fails to build.

Project created with only 'React' works ok.

— Looks like it's because of Typescript definitions? Is there a workaround?

Thanks!

Details:
OSX 10.11.6
Node v6.11.0
Installed dotnet-dev-osx-x64.1.0.4.pkg

Reproduce:
```
$ npm install -g yo generator-aspnetcore-spa

$ mkdir test1
$ cd test1
$ yo aspnetcore-spa
// choose React with Redux
// web pack fails
// full output:

     _-----_     ╭──────────────────────────╮
    |       |    │  Welcome to the ASP.NET  │
    |--(o)--|    │   Core Single-Page App   │
   `---------´   │        generator!        │
    ( _´U`_ )    │                          │
    /___A___\   /│      Version: 0.9.3      │
     |  ~  |     ╰──────────────────────────╯
   __'.___.'__   
 ´   `  |° ´ Y ` 

? Framework React with Redux
? Your project name test1


I'm all done. Running npm install for you to install the required dependencies. If this fails, try running the command yourself.


   create package.json
   create .template.config/icon.png
   create appsettings.json
   create ClientApp/boot-client.tsx
   create ClientApp/boot-server.tsx
   create ClientApp/components/Counter.tsx
   create ClientApp/components/FetchData.tsx
   create ClientApp/components/Home.tsx
   create ClientApp/components/Layout.tsx
   create ClientApp/components/NavMenu.tsx
   create ClientApp/configureStore.ts
   create ClientApp/css/site.css
   create ClientApp/dist/89889688147bd7575d6327160d64e760.svg
   create ClientApp/dist/main-server.js
   create ClientApp/dist/vendor-manifest.json
   create ClientApp/dist/vendor.js
   create ClientApp/routes.tsx
   create ClientApp/store/Counter.ts
   create ClientApp/store/index.ts
   create ClientApp/store/WeatherForecasts.ts
   create Controllers/HomeController.cs
   create Controllers/SampleDataController.cs
   create global.json
   create Program.cs
   create Startup.cs
   create .gitignore
   create Test1.csproj
   create tsconfig.json
   create Views/_ViewImports.cshtml
   create Views/_ViewStart.cshtml
   create Views/Home/Index.cshtml
   create Views/Shared/_Layout.cshtml
   create Views/Shared/Error.cshtml
[...]

PR #1067, opened 2017-06-26
newer version react-router 4.0.12 breaks - ugly force version 4.0.11 #1066

eino-makitalo:
Newer version of react-router 4.0.12 breaks typescript compilation of template reactredux  Probably needs some better fix but this is a workaround.


SteveSandersonMS:
Thanks for this! I'm just in the process of making a broader set of changes that will pin all dependency versions for long-term stability, so will cover this as part of that. Hope that's OK!

PR #1091, opened 2017-07-06
Set "TreatWarningsAsErrors" before NuGet restore

ryanbrandenburg:
* Ensures our build stays clean of NuGet warnings


ryanbrandenburg:
Merging since this passes locally.

PR #1102, opened 2017-07-11
On Windows, patch Node's lstat to work around permissions error. Fixes #1101

### LINKED ISSUE #1101: Error "EPERM: operation not permitted, lstat ..." when deployed to full IIS with restricted permissions
In a typical IIS deployment, the application runs as a user that only has permission to read the filesystem at the level of the application and below. It does *not* usually have permission to read any attributes of ancestor-level directories.

This causes Node to throw an error when it tries to load any module. For example, if the module file was at `C:\Users\username\apps\myapp1\`, the IIS user would typically have access to read attributes of `C:\` and `C:\Users`, but *not* `C:\Users\username`, so Node would throw:

```
Error: EPERM, operation not permitted 'C:\Users\username'
    at Error (native)
    at Object.bindingFs.lstat ([eval]:1:138)
    at Object.fs.lstatSync (fs.js:796:18)
    at Object.realpathSync (fs.js:1387:21)
    at tryFile (module.js:135:15)
    at Function.Module._findPath (module.js:174:18)
    at Function.Module._resolveFilename (module.js:334:25)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
```

The underlying issue is [a bug which was already reported to Node](https://github.com/nodejs/node-v0.x-archive/issues/3977), i.e., that Node tries to walk along the file path to the module, starting from the disk root, reading the attributes of each directory to see whether or not it's a symlink. It's fine to do that in typical Linux/macOS cases, but is not at all fine in typical Windows cases.

This affects all uses of NodeServices (both the SPA templates, and any standalone usage) for applications deployed to IIS, such as those deployed to a local IIS instance by Visual Studio.

SteveSandersonMS:
It's not great to have to patch Node internals like this, but it's the best we've got in the short term. See #1101 for details of the problem this solves.

This logic was tested on Node versions 0.12.0, 4.6.2, 6.11.0, and 8.1.3 (Windows 10, 64 bit).

Longer term, we need to work with the Node folks to get a proper fix into Node itself and then make this patching logic conditional on older versions of Node.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ install:
    - ps: Install-Product node 6.9.2 x64
    # .NET Core SDK binaries
    # Download .NET Core 2.0 Preview 3 SDK and add to PATH
-   - ps: $urlCurrent = "https://dotnetcli.azureedge.net/dotnet/Sdk/2.0.0-preview3-006670/dotnet-sdk-2.0.0-preview3-006670-win-x64.zip"
+   - ps: $urlCurrent = "https://dotnetcli.azureedge.net/dotnet/Sdk/2.0.0-preview3-006729/dotnet-sdk-2.0.0-preview3-006729-win-x64.zip"
    - ps: $env:DOTNET_INSTALL_DIR = "$pwd\.dotnetsdk"
    - ps: mkdir $env:DOTNET_INSTALL_DIR -Force | Out-Null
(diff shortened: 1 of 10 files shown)


SteveSandersonMS:
Approved by @Eilon

deva11:
downgrade npm version to 5.4.2 .worked for me!

PR #1107, opened 2017-07-11
Added npm Install on first-run

kirkone:
on my build agent the build failed because npm install was not running bevor the build. I can not see why npm install should only run when publishing.


stephtr:
Maybe it would be useful to run ```npm restore``` only when the ```node_modules``` folder isn't present (even though the time saving is quite low).

android2221:
What about the folks using yarn? I could say for myself that I would not be excited if this ran the install for me. I agree with the fact that this isn't much of a time-saver.

kirkone:
I close this because it is open for too long and I think this will not be merged.

PR #1113, opened 2017-07-12
Update templates to support TypeScript's strict mode

### LINKED ISSUE #1098: Changes required for compiling in strict mode
**EDIT:** This issue has nothing to do with projects generated using the template throwing errors while running webpack. For a solution to that problem see [#1099](https://github.com/aspnet/JavaScriptServices/issues/1099#issuecomment-313876245).

---

Even though I'm not sure if TypeScript's ```strict``` should be activated by default, I think that the templates should at least compile when a user decides to enable strict mode.
At the moment the ReactRedux template throws a few issues:

- ```./ClientApp/boot-client.tsx:5:30 TS7016: Could not find a declaration file for module 'react-hot-loader'.```

- ```./ClientApp/configureStore.ts:15:51 TS7006: Parameter 'f' implicitly has an 'any' type.```

- ```./ClientApp/configureStore.ts:33:27 TS7006: Parameter 'allReducers' implicitly has an 'any' type.```

- ```./ClientApp/store/WeatherForecasts.ts:63:7 TS2322: Type '{ startDateIndex: null; forecasts: never[]; isLoading: false; }' is not assignable to type 'WeatherForecastsState'. Types of property 'startDateIndex' are incompatible. Type 'null' is not assignable to type 'number'.```

All of these issues can be fixed with just small modifications. Should I create a PR?

stephtr:
Would fix #1098 and all other templates.
Additionally I switched from `requirejs` and `node` definitions to `webpack-env`, which was used by the React templates and I think is the correct solution.
The Angular template still has an (external) error [angular/#17863](https://github.com/angular/angular/issues/17863) remaining (only when compiling with strict flag).


templates/AngularSpa/ClientApp/boot.browser.ts
```diff
@@ -5,13 +5,13 @@ import { enableProdMode } from '@angular/core';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { AppModule } from './app/app.module.browser';
 
-if (module['hot']) {
-    module['hot'].accept();
-    module['hot'].dispose(() => {
+if (module.hot) {
```
stephtr: used node's definition, switched to webpack's

templates/ReactReduxSpa/ClientApp/boot-client.tsx
```diff
@@ -12,7 +12,7 @@ import * as RoutesModule from './routes';
 let routes = RoutesModule.routes;
 
 // Create browser history to use in the Redux store
-const baseUrl = document.getElementsByTagName('base')[0].getAttribute('href');
+const baseUrl = document.getElementsByTagName('base')[0].getAttribute('href')!;
```
stephtr: When we assume that a base element exists, we can also assume that it has the `href` attribute set.

templates/ReactSpa/ClientApp/boot.tsx
```diff
@@ -10,7 +10,7 @@ let routes = RoutesModule.routes;
 function renderApp() {
     // This code starts up the React app when it runs in a browser. It sets up the routing
     // configuration and injects the app into a DOM element.
-    const baseUrl = document.getElementsByTagName('base')[0].getAttribute('href');
+    const baseUrl = document.getElementsByTagName('base')[0].getAttribute('href')!;
```
stephtr: When we assume that a base element exists, we can also assume that it has the `href` attribute set.

templates/ReactReduxSpa/ClientApp/components/FetchData.tsx
```diff
@@ -56,8 +56,8 @@ class FetchData extends React.Component<WeatherForecastProps, {}> {
     }
 
     private renderPagination() {
-        let prevStartDateIndex = this.props.startDateIndex - 5;
-        let nextStartDateIndex = this.props.startDateIndex + 5;
+        let prevStartDateIndex = (this.props.startDateIndex || 0) - 5;
```
stephtr: Usually `startDateIndex` gets set in `componentWillMount` by calling `requestWeatherForecasts`.

templates/AngularSpa/ClientApp/boot.server.ts
```diff
@@ -17,12 +17,12 @@ export default createServerRenderer(params => {
     ];
 
     return platformDynamicServer(providers).bootstrapModule(AppModule).then(moduleRef => {
-        const appRef = moduleRef.injector.get(ApplicationRef);
+        const appRef: ApplicationRef = moduleRef.injector.get(ApplicationRef);
```
stephtr: Probably because `ApplicationRef` is an abstract class this type wasn't inferred automatically.

SteveSandersonMS:
Thanks very much! This is great. It's now merged.

I also turned on `strict` mode for all the templates. In the Angular case, I also had to add `skipLibCheck: true` because of the Angular type definitions bug you mentioned. We can remove that line once Angular fixes the issue.

PR #1128, opened 2017-07-19
Remove unused variable AutoMapperVersion

mikeharder:
@Eilon: `2.0` or `dev`?


SteveSandersonMS:
Thanks! I've merged this into `dev` because it won't make any difference to put it in `rel/2.0.0`, and we're trying to minimise the amount of commit noise on that branch.

PR #1161, opened 2017-07-31
fix(webpack): fix middleware to specifically serialize options as pascal case

stephenlautier:
This fixes middleware when consumers override serializer defaults settings via `JsonConvert.DefaultSettings` which would break webpack options passed to node. 

Since it's expected to be as Pascal case, it's safer to specifically specify it, and in our case we had it default to camelCase which was silently breaking it (and it was quite hard to figure it out).


stephenlautier:
If you agree with the PR, can someone look into why the build is failing, as it doesnt look it has to do anything with my changes. Thanks

SteveSandersonMS:
Thanks @stephenlautier!

Your change looks good. The AppVeyor build fail on the `dev` branch is a temporary issue until the final ASP.NET Core 2.0 packages are pushed to the public NuGet feed. We have another internal CI system that is still passing, so I know it's still OK.

PR #1162, opened 2017-08-01
Fix typo

chlowell:
Removes an extraneous "is" from the Angular template app's home component.


SteveSandersonMS:
Thanks @chlowell!

PR #1163, opened 2017-08-01
HttpNodeInstance - Added Timeout to HttpClient to always be longer than the InvocationTimeoutMilliseconds

ericgreenmix:
HttpClient defaults to 100 seconds. If InvocationTimeoutMilliseconds is greater than the default of the HttpClient, the task will fail with a Task was canceled exception.


SteveSandersonMS:
Thanks for this @ericgreenmix!

PR #1169, opened 2017-08-03
Fix problems with AoT when using Angular Material. Fixes #1168

### LINKED ISSUE #1168: Angular AOT error when having a component within a library
Hi, 

Not sure if this an issue, or requires some changes within webpack in order to get it work, perhaps someone else already seen this issue as it should be quite common I would say. 
When adding a library such as Angular Material, or even I created a simple component within a library to narrow this down, and building the app as AoT, we get the following error:

```bash
    ERROR in ./$$_gendir/~/@odin/ngx.core/dist/xyz.component.ngfactory.ts
    Module parse failed: C:\dev\git\loki\$$_gendir\node_modules\@odin\ngx.core\dist\xyz.component.ngfactory.ts Unexpected token (11:25)
    You may need an appropriate loader to handle this file type.
    | import * as i0 from '@angular/core';
    | import * as i1 from '@odin/ngx.core/dist/xyz.component';
    | const styles_XyzComponent:any[] = ([] as any[]);
    | export const RenderType_XyzComponent:i0.RendererType2 = i0.ɵcrt({encapsulation:2,styles:styles_XyzComponent,
    |     data:{}});
     @ ./$$_gendir/ClientApp/app/areas/home/home.component.ngfactory.ts 9:0-104
     @ ./$$_gendir/ClientApp/app/app.module.browser.ngfactory.ts
     @ ./ClientApp/boot-browser.ts
```
The following is a simple component which doesn't work

```ts
import { Component } from "@angular/core";

@Component({
	selector: "odn-xyz",
	template: `<h1>Xyz5</h1> <button>a button</button>`
})
export class XyzComponent {

}
```

These are the conditions when it doesn't work.
* when built with aot (without AoT it works fine)
* and in a separate library (as a vendor)
* and library contains a component

SteveSandersonMS:


SteveSandersonMS:
This was merged into `dev` instead.

PR #1217, opened 2017-08-22
Added ability to load files relative to the app root

### LINKED ISSUE #1205: ReactRedux configuration doesn't play well with TypeScript tools
Current configuration of ReactRedux project doesn't work well with TypeScript code actions. In my case I have two files:
* /ClientApp/logs/components/LogsList.tsx
* /ClientApp/logs/components/OperationLogs.tsx

When I try to add import to _LogsList.tsx_ from _OperationLogs.tsx_ the following line is added:

    import { LogsList } from 'ClientApp/logs/components/LogsList';

Visual Studio doesn't show any errors but then when I run application I get such error:

NodeInvocationException: Prerendering failed because of error: Error: Cannot find module "ClientApp/logs/components/LogsList"

I have already reported to TypeScript repository that in such case they could use relative path https://github.com/Microsoft/TypeScript/issues/15888#issuecomment-323339210, but it would be nice to have that correctly configured in template.

rosieks:
Fix #1205


SteveSandersonMS:
Thanks for this, but could you clarify why you think we should make this change?

It looks like you're redefining `ClientApp` as the app root, rather than what is usually considered to be the app root, i.e., the directory containing the `.csproj`, `Startup.cs`, etc. That could also be confusing, because then there are effectively two app root directories, depending on what kind of technology you're talking about. If possible I'd prefer to keep the regular app root as being the only app root.

rosieks:
The main change is to add any root path in webpack configuration. Right now such root is no provided so when I add in any file such code `import { reducer } from 'ClientApp/store';` Visual Studio and TypeScript will be satisfied because of `"rootUrl": "."`, but webpack doesn't know that and it fails during execution. For that reason I have added modules section in webpack.config.js. Detailed description on that change you can find here: http://moduscreate.com/es6-es2015-import-no-relative-path-webpack/

Regarding setting app root to ClientApi I did that because de facto it is the app root from JavaScript perspective. I'm not sure if somebody reference from it to something outside. But I can be wrong and it can be changed.

SteveSandersonMS:
Thanks again for posting this.

In the next version of the ReactRedux template, the build config comes entirely from create-react-app. There won't be a Webpack config file, so it would be impossible to apply this PR. As such, I'll mark this closed. Hope that's OK!

PR #1235, opened 2017-08-29
Use Directory.Build.props/targets

natemcmaster:
We've started using Directory.Build.props/targets across aspnet repos to reduce the duplication between csproj files, and reduce the amount of boilerplate code required to add new projects and tests.


src/Microsoft.AspNetCore.NodeServices/Microsoft.AspNetCore.NodeServices.csproj
```diff
@@ -1,29 +1,14 @@
...
   <PropertyGroup>
     <Description>Invoke Node.js modules at runtime in ASP.NET Core applications.</Description>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <PackageTags>aspnetcore;aspnetcoremvc;nodeservices</PackageTags>
-    <TypeScriptCompileBlocked>true</TypeScriptCompileBlocked>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
-    <None Remove="node_modules\**\*" />
```
SteveSandersonMS: I'm not keen on this or any of the following lines in this `csproj` being moved out into `Directory.Build.props`, because:

1. It's hard to reason about which `node_modules` or `Content` files are being included/excluded when the paths aren't even relative to the file that's referencing them
2. It makes an assumption about conventions for inclusion/exclusion and prepublishing always being the same across every project in `src`, which isn't guaranteed. Although they might all happen to follow the same conventions today, there's no reason to believe that should always be so. Not all of them should necessarily have a `Content` directory, use `node_modules`, etc.

So I would prefer not to deduplicate these values, but rather keep the `src` projects equivalently decoupled as they were before.

natemcmaster: Sounds good. I'll revert this piece of the change.

natemcmaster:
:up: :date:

natemcmaster:
I think I address your feedback, so I'm going to merge this so I can get #1236 in too. If there was anything I missed, let me know and I'll fix it.

PR #1243, opened 2017-08-30
Pack and test templates from MSBuild

natemcmaster:
Moves build/test scripting from appveyor.yml into repo.targets.


SteveSandersonMS:
Thanks @natemcmaster for this. When you suggested moving the AppVeyor steps to a new `.targets` file, I didn't realise this would involve a whole bunch of work for you.

As it happens we're imminently moving the SPA templates out of this repo and into the `aspnet/templating` repo, which is already set up to build/pack/test the templates via MSBuild. So, rather than merge this PR, we should obtain mostly the same improvements automatically by virtue of moving the template content across.

One thing we won't get automatically in `aspnet/templating` is the browser automation system. I'll be adding something like that in that repo very soon, and will probably benefit from your work here - I can probably copy some of your implementation (if we're sticking with a Selenium-based system rather than moving to something purely in .NET).

So if it's OK with you I'll leave this PR open for a few days while I use it for reference, but will then close without merging once the SPA templates are moved out of this repo. Does that sound OK?

natemcmaster:
> So if it's OK with you I'll leave this PR open for a few days while I use it for reference, but will then close without merging once the SPA templates are moved out of this repo. Does that sound OK?

Sounds good to me. I'll close it now, but will leave the namc/korebuild branch until you're done with it. 

FYI - we should still look into removing custom code from appveyor.yml. This file is currently installing a stale version of the SDK. IMO it would be better to use KoreBuild to install the .NET Core SDK instead of having to keep appveyor.yml up to date.

https://github.com/aspnet/JavaScriptServices/blob/78436adb08a36f9ba6fac5c398fa0da9af38c010/.appveyor.yml#L7

PR #1246, opened 2017-09-01
Update comment for large screen media query

BaylorRae:
Just found this project today and so far it has been great to work with. Thanks for all the work put into making these Visual Studio templates.

When going through the CSS I noticed this typo for the desktop media query comments. This PR fixes the comments in the template css files replaces **small** with **large** where appropriate.


BaylorRae:
These files were moved to [aspnet/templating](https://github.com/aspnet/templating) in 09317b83a86

PR #1250, opened 2017-09-02
redux template: automatically infer FetchData's new props type

stephtr:
After being connected to redux's store the props type isn't the same anymore since `WeatherForecastsState` gets filled by the store.

The template worked because when using the `FetchData` component within the `Route` component it seems like its type doesn't get checked accurately. However using the component as `<FetchData ... />` generates an error since the properties of `WeatherForecastsState` were seemingly missing.

The type definition of redux's `connect` has been built in such a way, that the new props type gets automatically infered.


SteveSandersonMS:
Thanks for submitting this PR! As it happens, the next version of the React-Redux template will be based around create-react-app output and this change won't be applicable to it (nor will the original issue). I'll close this because it doesn't apply to the newer template. Hope that's OK!

stephtr:
That is nice to hear!

PR #1252, opened 2017-09-02
Returning provided promise in `addTask`

### LINKED ISSUE #1249: [domain-task] addTask: return promise
Wouldn't it be more useful if `addTask` would return the original promise?
One could then write `const response = await addTask(fetch('/api/test'));` instead of
```
const task = fetch('/api/test');
addTask(task);
const response = await task;
```
Edit: [@DanHarman](https://github.com/aspnet/JavaScriptServices/issues/902#issuecomment-299418722) also mentioned it.

rosieks:
This PR resolves #1249


SteveSandersonMS:
Thanks!

PR #1284, opened 2017-09-18
[2.0.1] Update references to other 2.0.1 patch packages

natemcmaster:
- Adding the patch feed
 - Adding the generated list of packages that should be changed from AspNetCoreVersion


natemcmaster:
FYI - CI is expected to fail until we have an initial build

PR #1306, opened 2017-09-29
Add SPA middleware APIs to support new templates

SteveSandersonMS:
These new APIs allow for a more modern way to host a SPA inside an ASP.NET application:

* Instead of prerendering being done through a tag helper that executes when rendering an MVC view, prerendering can now be applied as root-level middleware independent of any MVC concepts.
  * This means it works just the same in Razor Pages sites as in MVC sites.
  * In this mode, the prerenderer is expected to return a complete HTML page rather than just an HTML fragment, which better matches what people want to do with more recent SPA technologies such as Angular CLI.
* Instead of SPA fallback routing being done as an MVC route, it can now be applied as another root-level middleware that returns a "default page" for the SPA.
  * This is simpler and again it better matches more recent SPA technologies.
  * Plus it allows a new notion of SPA-related middleware that specifically modify how the "default page" is generated - for example it might be generated by Webpack Dev Middleware or Angular CLI.
* The dev-time middleware is no longer limited to Webpack. This PR also adds support for Angular CLI to act as alternative dev-time middleware to return dynamically built client-side resources.

None of these changes should affect backward compatibility. Existing SPA applications built with the current templates can continue to work fine and don't need to be updated.


SteveSandersonMS:
I'm going to merge this straight away because some other template work can't proceed until we get CI builds of the updated `SpaServices` package.

But this is still very much open to review and possible changes/reversions.

src/Microsoft.AspNetCore.SpaServices/AngularCli/AngularCliBuilder.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <inheritdoc />
+        public Task Build(ISpaBuilder spaBuilder)
+        {
+            // Locate the AngularCliMiddleware within the provided ISpaBuilder
+            var angularCliMiddleware = spaBuilder
+                .Properties.Keys.OfType<AngularCliMiddleware>().FirstOrDefault();
+            if (angularCliMiddleware == null)
+            {
+                throw new Exception(
```
rosieks: Framework Design Guidelines: DO NOT throw Exception or SystemException.

src/Microsoft.AspNetCore.SpaServices/Prerendering/SpaPrerenderingExtensions.cs
```diff
@@ -0,0 +1,126 @@
...
+            {
+                context.Response.Redirect(renderResult.RedirectUrl);
+            }
+            else
+            {
+                // The Globals property exists for back-compatibility but is meaningless
+                // for prerendering that returns complete HTML pages
+                if (renderResult.Globals != null)
+                {
+                    throw new Exception($"{nameof(renderResult.Globals)} is not supported when prerendering via {nameof(UsePrerendering)}(). Instead, your prerendering logic should return a complete HTML page, in which you embed any information you wish to return to the client.");
```
rosieks: Framework Design Guidelines: DO NOT throw Exception or SystemException.

src/Microsoft.AspNetCore.SpaServices/SpaExtensions.cs
```diff
@@ -0,0 +1,120 @@
...
+                    // Try to clarify the common scenario where someone runs an application in
+                    // Production environment without first publishing the whole application
+                    // or at least building the SPA.
+                    var hostEnvironment = (IHostingEnvironment)context.RequestServices.GetService(typeof(IHostingEnvironment));
+                    if (hostEnvironment != null && hostEnvironment.IsProduction())
+                    {
+                        message += "Your application is running in Production mode, so make sure it has been published, or that you have built your SPA manually. Alternatively you may wish to switch to the Development environment.\n";
+                    }
+
+                    throw new Exception(message);
```
rosieks: Framework Design Guidelines: DO NOT throw Exception or SystemException.

PR #1311, opened 2017-10-02
Use built-in GetRequiredService<T>

khellang:
...which uses [the `ISupportRequiredService` optimization](https://github.com/aspnet/DependencyInjection/blob/88c3bd6fe2786dd759b4a6c6d7c410e895336b6c/src/DI.Abstractions/ServiceProviderServiceExtensions.cs#L50-L54)

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices/Prerendering/SpaPrerenderingExtensions.cs
+++ b/src/Microsoft.AspNetCore.SpaServices/Prerendering/SpaPrerenderingExtensions.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.NodeServices;
 using Microsoft.AspNetCore.SpaServices;
 using Microsoft.AspNetCore.SpaServices.Prerendering;
+using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Threading.Tasks;
 
@@ -41,9 +42,9 @@ namespace Microsoft.AspNetCore.Builder
             var appBuilder = spaBuilder.AppBuilder;
             var serviceProvider = appBuilder.ApplicationServices;
             var nodeServices = GetNodeServices(serviceProvider);
-            var applicationStoppingToken = GetRequiredService<IApplicationLifetime>(serviceProvider)
+            var applicationStoppingToken = serviceProvider.GetRequiredService<IApplicationLifetime>()
                 .ApplicationStopping;
-            var applicationBasePath = GetRequiredService<IHostingEnvironment>(serviceProvider)
+            var applicationBasePath = serviceProvider.GetRequiredService<IHostingEnvironment>()
                 .ContentRootPath;
             var moduleExport = new JavaScriptModuleExport(entryPoint);
 
@@ -90,12 +91,6 @@ namespace Microsoft.AspNetCore.Builder
 
                 await ApplyRenderResult(context, renderResult);
             });
-        }
-
-        private static T GetRequiredService<T>(IServiceProvider serviceProvider) where T: class
-        {
-            return (T)serviceProvider.GetService(typeof(T))
-                ?? throw new Exception($"Could not resolve service of type {typeof(T).FullName} in service provider.");
         }
 
         private static async Task ApplyRenderResult(HttpContext context, RenderToStringResult renderResult)


SteveSandersonMS:
Thanks - that's better!

khellang:
@SteveSandersonMS Did you get to take a look at https://gist.github.com/khellang/47e8fac5b1bad2df74cb8c145d32f98e? Would that be something to add to SpaServices?

SteveSandersonMS:
I'm considering it. We have something very similar going into SpaServices for 2.1 whereby we replace the functionality formerly in `MapSpaFallbackRoute` with a new root-level middleware (see: [early implementation, subject to much change](https://github.com/aspnet/JavaScriptServices/blob/dev/src/Microsoft.AspNetCore.SpaServices/SpaExtensions.cs#L43) - this also uses a marker, but does so to flag which requests have had their paths rewritten).

Possibly something in your approach would help with that, but I haven't determined that yet. I'm particularly interested in whether your approach to "re-run the middleware chain" might make this more flexible and avoid the need to chain on a second copy of `UseStaticFiles` internally.

khellang:
> I'm particularly interested in whether your approach to "re-run the middleware chain" might make this more flexible and avoid the need to chain on a second copy of `UseStaticFiles` internally.

Yeah, avoiding two instances of the StaticFiles MW was one goal. It's pretty flexible in terms of where you place the fallback MW, i.e. how much of your pipeline you want to re-execute, and who ends up serving the actual fallback. It could be StaticFiles, it could be MVC, or something entirely different.

SteveSandersonMS:
Thanks for the suggestions. I've been circling around various approaches and currently here's how it's looking: https://github.com/aspnet/templating/blob/feature/stevesa/angular-cli-template/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/Startup.cs#L41, with implementation called by `UseSpa` here: https://github.com/aspnet/JavaScriptServices/blob/dev/src/Microsoft.AspNetCore.SpaServices/SpaDefaultPageMiddleware.cs#L22

I realised that we wouldn't get away with the approach of running the same middleware chain more than once. It does work nicely in simple scenarios, but crazy things would happen if somebody (for example) added middleware that did logging or had other side effects. They wouldn't expect it to fire more than once per request.

So instead it does have to add a second `StaticFiles` middleware specifically to serve the `index.html`. I can't think of any problems this will cause, or any perf drawbacks, but please let me know if you are aware of any.

khellang:
> I realised that we wouldn't get away with the approach of running the same middleware chain more than once. It does work nicely in simple scenarios, but crazy things would happen if somebody (for example) added middleware that did logging or had other side effects. They wouldn't expect it to fire more than once per request. Debugging would be super confusing too.

Hmm... 🤔 I'd be interested in some example scenarios where you think this would be a problem. Is this a pattern that would generally be recommended against?

The way I (and the MW in the pipeline) see it, it's just a "server-side redirect", i.e. two different requests: one for the client-requested path, one for the fallback path. I've even thought about setting a different trace identifier for the fallback, but figured it could be worth keeping it to correlate which paths were requested and ended up serving the fallback.

khellang:
Anyway, it's been working pretty well for me (for a couple of fairly complex apps), but I've been holding off shipping it as a NuGet package in case you wanted to roll some of it into SpaServices. I guess this is my cue to go ahead and finally package it up and ship it :shipit: 😉

SteveSandersonMS:
> I'd be interested in some example scenarios where you think this would be a problem.

A simple middleware like:

```
app.use(async (context, next) => {
    // Set breakpoint on next line:
    Console.WriteLine("Received request for " + context.Request.Path);
});
```

...would be weird because:

1. It logs twice per request
2. Breakpoint hits twice per request

I judged that adding the extra StaticFiles middleware was less of a problem (in fact, not a problem at all, AFAICT).

khellang:
> Is this a pattern that would generally be recommended against?

I just realized that there's precedent from MS on this already. The StatusCodePages MW has an option to re-execute the pipeline 😊

> I judged that adding the extra StaticFiles middleware was less of a problem (in fact, not a problem at all, AFAICT).

Yeah, I don't think there will be any problems with it. The reason I didn't want two instances is because in some apps, I have some custom settings for the static files MW. When adding another instance, I would have to provide another overload for my MW, taking in static file options. The other thing is simply that I'd like to avoid the coupling to the static files MW. In my case, I could just as easily serve the fallback from any other MW, because it's just a path :smile:

PR #1341, opened 2017-10-16
Add RepositoryRoot

ryanbrandenburg:


SteveSandersonMS:
@ryanbrandenburg I'm sure you're adding this for a good reason, but I don't know what it is. Would be great for these kinds of PRs to come with a 1-sentence description of their purpose.

ryanbrandenburg:
I recently added SourceLink to KoreBuild, which uses the RepositoryRoot property as part of its input. KoreBuild calculates that property automatically, but VS doesn't. Because of this when projects were being run in VS they would throw a warning for every project warning you that SourceLink wasn't running.

I will look at updating the script I use to update every repo with the ability to provide a description.

SteveSandersonMS:
Thanks Ryan!

PR #1371, opened 2017-11-02
Added a more updated alternative to isomorphic-fetch.

lquixada:
isomorphic-fetch hasn't been updated for a while, so cross-fetch was created in order to provide a more updated, flexible and [fixed](https://github.com/matthew-andrews/isomorphic-fetch/issues/125) alternative to the community.


SteveSandersonMS:
Thanks for contributing this.

Could you describe what's not working well enough about `isomorphic-fetch`, preferably giving an example of a problem that occurs when it gets used by `domain-task`? It's not that there's any reason to avoid `cross-fetch`, but rather that since this is an internal dependency, it only makes sense to change it if there's an externally observable problem.

SteveSandersonMS:
I'll close this due to lack of information, but please feel free to return with more details if you think it's appropriate. Thanks!

PR #1372, opened 2017-11-03
Update appveyor to run korebuild

natemcmaster:
AppVeyor isn't actually building the repo because the default AppVeyor image is VS 2015, not 2017. 

Changes:
 - update appveyor to vs2017
 - execute a default build with korebuild
 - add 'nodejs' as a required toolset in korebuild.json. This ensures nodejs is available on the machine. (Assuming this is required so tests can run)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,19 +1,21 @@
-﻿init:
+init:
   - git config --global core.autocrlf true
 install:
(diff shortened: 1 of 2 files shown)


natemcmaster:
Yup, CI is green :) https://ci.appveyor.com/project/aspnetci/javascriptservices/build/1.0.573

PR #1547, opened 2018-02-21
Fix #1505

### LINKED ISSUE #1505: UseSpaStaticFiles and UserSpa should return with IApplicationBuilder
UseSpaStaticFiles and UseSpa extension methods are void and not return with  IApplicationBuilder as any other AppBuilder extension method in Asp.Net Core

Is there any reason why they are void? If there is no, then it would be great to return with IApplicationBuilder.

Thanks

pherbel:
Fix for #1505

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/Configuration/NodeServicesServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNetCore.NodeServices/Configuration/NodeServicesServiceCollectionExtensions.cs
@@ -12,6 +12,7 @@ namespace Microsoft.Extensions.DependencyInjection
         /// Adds NodeServices support to the <paramref name="serviceCollection"/>.
         /// </summary>
         /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
+        /// <returns>A reference to this instance after the operation has completed.</returns>
         public static IServiceCollection AddNodeServices(this IServiceCollection serviceCollection)
             => AddNodeServices(serviceCollection, _ => {});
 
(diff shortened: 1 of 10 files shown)


pherbel:
@SteveSandersonMS 
Could you please review it and merge if it possible?
Could it be part of the next preview release?
Thanks

SteveSandersonMS:
This looks good - thanks for contributing it, @pherbel!

The only thing missing is updates to the XML docs. If a `public` method changes from `void` to non-`void`, then in its XML docs we also need to add a `<returns>...</returns>` entry to document what it now returns.

Example:

```csharp
        /// <summary>
        /// Adds NodeServices support to the <paramref name="serviceCollection"/>.
        /// </summary>
        /// <param name="serviceCollection">The <see cref="IServiceCollection"/>.</param>
        /// <returns>A reference to this instance after the operation has completed.</returns>
        public static IServiceCollection AddNodeServices(this IServiceCollection serviceCollection)
            => AddNodeServices(serviceCollection, _ => {});
```

In every case it looks like you're returning the `this` value, so the description *A reference to this instance after the operation has completed.* would be OK for all of them.

Would you mind updating the XML docs in that way? Then I'd definitely be happy to merge it.

pherbel:
Thanks @SteveSandersonMS 

It absolutely makes sense. I missed it first time.
I fixed the missing xml docs and I also added return value and xml doc for other builder extension method what I forgot too. Now it will be consistent in all builder API.

Thanks your time!

Eilon:

Thanks,
Eilon

Issue #1361, opened 2017-10-28
Webpack 3 compatibility


michaldudak:
Please consider making the `aspnet-webpack` package (and others such as `aspnet-webpack-react`) compatible with webpack 3. Currently yarn issues the following warnings:

```
warning "aspnet-webpack@2.0.1" has incorrect peer dependency "webpack@^1.13.2 || ^2.1.0-beta".
warning "aspnet-webpack-react@3.0.0" has incorrect peer dependency "webpack@^2.2.0".
```

However, despite these warnings, the packages seem to be working fine (although I have not tested them extensively).

MovGP0:
confirmed. I have the same issue
```
warning "aspnet-webpack@2.0.1" has incorrect peer dependency "webpack@^1.13.2 || ^2.1.0-beta".
```

MovGP0:
apparently the support for Webpack 3.0 does not work perfectly yet. When the server is still loading and a request already made, the request will be stuck waiting and canceled after the timeout of the cancellation token. 

After the server has fully loaded, additional requests will work fine. 

```
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()
   at System.Net.Http.DiagnosticsHandler.<SendAsync>d__2.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
```
(snipped stacktrace for clarity) 
```
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.SpaServices.Webpack.ConditionalProxyMiddleware.<Invoke>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #1559, opened 2018-03-02
Update peerDependency to support webpack 3

jafin:
Update peerDependency to support webpack 3.
- This is only to remove the npm peer dep warning.
- Reports of various people already running 3.x are listed in.

Addresses #1361


GeekyMonkey:
Why stop there. Webpack 4 is released now.

jafin:
@GeekyMonkey #1361 Issue was for Webpack3.  Mayby you could open an issue for Webpack 4 and PR.  We don't have Webpack 4 here so I wouldn't be able to test sorry.

SteveSandersonMS:
Fixed with #1606. Thanks for contributing, @jafin!

Issue #1585, opened 2018-03-21
aspnet-webpack-react not compatible with react-hot-loader v4


Guymestef:
Hello,

react-hot-loader/patch has been removed from react-hot-loader v4: [See documentation here](https://github.com/gaearon/react-hot-loader#no-patch-required)

A new package for aspnet-webpack-react (v4.0.0?), removing support of "reactHotLoaderPatch", need to be created I suppose.

[To remove?](https://github.com/aspnet/JavaScriptServices/blob/04b8c17bdded1d7652f18587a920f740d42985ff/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/src/HotModuleReplacement.ts#L4)
```typescript
const reactHotLoaderPatch = 'react-hot-loader/patch';
```

[To remove?](https://github.com/aspnet/JavaScriptServices/blob/04b8c17bdded1d7652f18587a920f740d42985ff/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/src/HotModuleReplacement.ts#L52)
```typescript
        let entryValueArray = entryConfig[entrypointName] as string[];
        if (entryValueArray.indexOf(reactHotLoaderPatch) < 0) {
            entryValueArray.unshift(reactHotLoaderPatch);
        }
```

Guymestef:
For users still using react-hot-loader v3, they would be able to use this new version by adding react-hot-loader/patch to their webpack conf manually. No?

Guymestef:
By doing some test, it seems that the module react-hot-loader/webpack has been removed too...
```typescript
const reactHotLoaderWebpackLoader = 'react-hot-loader/webpack';
```

AndyCJ:
I'm getting the error about `'react-hot-loader/webpack'` being missing, so rolling back to V3 for now.

SteveSandersonMS:
> Would something like this work?

It looks like it would avoid the initial exception, but I'm not sure it solves the issue that HMR wouldn't work. Do you have an idea about how that should be changed for v4?

mark-monteiro:
In the migration instructions for migrating from v3 to v4, the only mention of changes to webpack config is the removal of `react-hot-loader/patch`: https://github.com/gaearon/react-hot-loader#migrating-from-v3

There is also a working example setup for HMR with Typescript that can be referenced: https://github.com/gaearon/react-hot-loader/tree/master/examples/typescript

[...]

PR #1675, opened 2018-06-07
Updating aspnet-webpack-react to work with react-hot-loader v4

tjaskula:
This should solve the issue #1585 

- Deleting references to `react-hot-loader/webpack` and `react-hot-loader/patch` as it was removed in v4 of react-hot-loader. Referencing `react-hot-loader/babel`in webpack configuration is enough.
- Bumping peer dependency of webpack to 3 and 4

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/package.json
+++ b/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aspnet-webpack-react",
-  "version": "3.0.0",
+  "version": "4.0.0",
   "description": "Helpers for using Webpack with React in ASP.NET Core projects. Works in conjunction with the Microsoft.AspNetCore.SpaServices NuGet package.",
   "main": "index.js",
   "scripts": {
@@ -17,12 +17,12 @@
     "url": "https://github.com/aspnet/JavaScriptServices.git"
   },
   "devDependencies": {
-    "@types/webpack": "^2.2.0",
+    "@types/webpack": "^4.4.0",
     "rimraf": "^2.5.4",
     "typescript": "^2.0.0",
-    "webpack": "^2.2.0"
+    "webpack": "^4.16.0"
   },
   "peerDependencies": {
-    "webpack": "^2.2.0 || ^3.0.0 || ^4.0.0"
+    "webpack": "^4.0.0"
   }
 }
--- a/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/package.json
+++ b/src/Microsoft.AspNetCore.SpaServices/npm/aspnet-webpack-react/package.json
@@ -23,6 +23,6 @@
     "webpack": "^2.2.0"
   },
   "peerDependencies": {
-    "webpack": "^2.2.0"
+    "webpack": "^2.2.0 || ^3.0.0 || ^4.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


natemcmaster:
Thanks for taking a look @SteveSandersonMS. 

Side note: I was playing around with this over the weekend at seemed to me like improvements to react-hot-loader and webpack 4 make aspnet-webpack-react unnecessary. I got hmr working just fine without it. I'm not proposing anything...just pointing out that I couldn't figure out what aspnet-webpack-react was supposed to do.

SteveSandersonMS:
@natemcmaster Were you using the SPA templates from ASP.NET Core 2.0 or later? If so, you're not using `aspnet-webpack-react` at all. It's only used by pre-2.0 templates.

natemcmaster:
I see, makes sense. I was setting up hmr with ASP.NET Core 2.1 on an old hobby project I was upgrading from ASP.NET 4. I didn't see HMR in the latest SPA templates so wasn't sure what to do.

tjaskula:
@SteveSandersonMS Thanks for suggestions. I'll target Webpack 4 only which will be related to the changes.
@natemcmaster You mean you made working HMR and ASP.NET core (on kestrel) without this module? I don't remeber now but it seems that enabling HMR yielded an error thus this PR.

natemcmaster:
> You mean you made working HMR and ASP.NET core (on kestrel) without this module?

@tjaskula yes. It may require updating to the latest versions of webpack and react-hot-loader, but I got it working just fine without aspnet-webpack-react. I wrote up a quick demo and post about this https://natemcmaster.com/blog/2018/07/05/aspnetcore-hmr/

tjaskula:
@natemcmaster I checked what was different in my configuration and realized that in ASP.NET Core I had this configuration:

```
app.UseWebpackDevMiddleware(new WebpackDevMiddlewareOptions
{
     HotModuleReplacement = true,
     ReactHotModuleReplacement = true,
     ProjectPath = System.IO.Path.Combine(env.ContentRootPath, "../")
});
```

Notice the flag `ReactHotModuleReplacement = true`. When this is set, the `aspnet-webpack-react` dependency is loaded with an error described in the issue #1585.

When I comment out `ReactHotModuleReplacement` the error goes away (I'm also using ASP.NET Core 2.1)

My point is that, if you're using ASP.NET Core 2.1, just go without `aspnet-webpack-react` and `ReactHotModuleReplacement` flag. (as @SteveSandersonMS mentioned it above)

For ASP.NET Core < 2.0 this PR might be of value.

I'm making changes to the PR as requested above.

cvanem:
@tjaskula Thanks for clearing this up.  Removing aspnet-webpack-react and REactHotMudleReplacement = true resolved my error and HMR still functions as expected.

tjaskula:
@natemcmaster do sourcemaps works for you ? I noticed that for me it doesn't load properly in Chrome

natemcmaster:
Yeah, I got those working with the inline-source-map devtool and awesome-typescript-loader.

tjaskula:
All is ok, I was using webpack.SourceMapdevtoolPlugin which seems broken. Swtiching to detool: inline-source-map seems fixing the issue.

tjaskula:
I've pushed the required changes. Please review if this is ok.

Tsourdox:
Really need this one. What's the eta for reviewing this PR?

Eilon:
@SteveSandersonMS ?

SteveSandersonMS:
Thanks for the prompt. This is now merged and published as `aspnet-webpack-react@4.0.0`.

PR #1679, opened 2018-06-09
Resolve relative Spa Source based on Content Root

### LINKED ISSUE #1678: ReactDevelopmentServer does not honor ContentRoot
I noticed that `ReactDevelopmentServer` starts a process with [working directory set to `SourcePath`](https://github.com/aspnet/JavaScriptServices/blob/96d7f853277788d52dc33027f31a8db4c4786cf1/src/Microsoft.AspNetCore.SpaServices.Extensions/Npm/NpmScriptRunner.cs#L56)  from `SpaOption` without considering that it may be relative, and thus should be relative to the [content root](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/index?view=aspnetcore-2.1&tabs=aspnetcore2x#content-root).

I think that it would make sense to append content root to `SourcePath` if `SourcePath` is relative. Then, as a bonus, I think it would be nice if an `DirectoryNotFoundException` was thrown if the source path doesn't exist. I got tricked by [generic error message](https://github.com/aspnet/JavaScriptServices/blob/96d7f853277788d52dc33027f31a8db4c4786cf1/src/Microsoft.AspNetCore.SpaServices.Extensions/Npm/NpmScriptRunner.cs#L122) when this failed for me and started to troubleshoot elsewhere.

pardahlman:
This PR closes #1678 by evaluating the provided `SourcePath` and doing the following:

1. Detect if path is relative and if so make it relative to the content root (with fallback to current directory)
2. Evaluate if the now absolute path exists and if not, throw a helpful exception
3. Using the existing, absolute path to start the `NpmScriptRunner`

### DIFF
--- a/src/Microsoft.AspNetCore.SpaServices.Extensions/ReactDevelopmentServer/ReactDevelopmentServerMiddleware.cs
+++ b/src/Microsoft.AspNetCore.SpaServices.Extensions/ReactDevelopmentServer/ReactDevelopmentServerMiddleware.cs
@@ -9,9 +9,12 @@ using Microsoft.AspNetCore.SpaServices.Util;
 using System;
 using System.IO;
 using System.Collections.Generic;
+using System.Runtime.InteropServices.ComTypes;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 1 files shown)


Eilon:

Thanks,
Eilon

PR #1725, opened 2018-07-18
Fix links that assumed branch "dev" exists (which doesn't)

MSeifert04:
I noticed that a lot of links in the readme actually lead to 404 pages because branch dev is missing (d742bea01e2d3179e142d708dc5be18ed2784432).

This just changes "dev" to "master", not sure if that's correct or desirable so feel free to close this again if not correct or necessary.

### DIFF
--- a/src/Microsoft.AspNetCore.NodeServices/README.md
+++ b/src/Microsoft.AspNetCore.NodeServices/README.md
@@ -10,7 +10,6 @@ This NuGet package provides a fast and robust way to invoke Node.js code from a 
 It is the underlying mechanism supporting the following packages:
 
  * [`Microsoft.AspNetCore.SpaServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.SpaServices) - builds on NodeServices, adding functionality commonly used in Single Page Applications, such as server-side prerendering, webpack middleware, and integration between server-side and client-side routing.
- * [`Microsoft.AspNetCore.AngularServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.AngularServices) and [`Microsoft.AspNetCore.ReactServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.ReactServices) - these build on `SpaServices`, adding helpers specific to Angular and React, such as cache priming and integrating server-side and client-side validation
 
 ### Requirements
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.NodeServices/README.md
```diff
@@ -9,8 +9,8 @@ This NuGet package provides a fast and robust way to invoke Node.js code from a
...
+ * [`Microsoft.AspNetCore.SpaServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.SpaServices) - builds on NodeServices, adding functionality commonly used in Single Page Applications, such as server-side prerendering, webpack middleware, and integration between server-side and client-side routing.
+ * [`Microsoft.AspNetCore.AngularServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.AngularServices) and [`Microsoft.AspNetCore.ReactServices`](https://github.com/aspnet/JavaScriptServices/tree/master/src/Microsoft.AspNetCore.ReactServices) - these build on `SpaServices`, adding helpers specific to Angular and React, such as cache priming and integrating server-side and client-side validation
```
MSeifert04: These links still don't work. Should I revert the change?

Eilon: You could remove the part that has the links that don't work (Angular and React services). Perhaps keep the rest?

src/Microsoft.AspNetCore.SpaServices/Webpack/ConditionalProxyMiddleware.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.AspNetCore.SpaServices.Webpack
 {
     /// <summary>
-    /// Based on https://github.com/aspnet/Proxy/blob/dev/src/Microsoft.AspNetCore.Proxy/ProxyMiddleware.cs
+    /// Based on https://github.com/aspnet/Proxy/blob/master/src/Microsoft.AspNetCore.Proxy/ProxyMiddleware.cs
```
Eilon: Might be best to shorten this to just `Based on ProxyMiddleware from https://github.com/aspnet/Proxy/.` that way we don't have to worry about adjusting this in the future.

Eilon:
@SteveSandersonMS - could you also take a look and then merge?

MSeifert04:
Thank you for the fast review 👍

PR #1745, opened 2018-08-23
Don't use depricated constructor

### LINKED ISSUE #1668: Buffer is deprecated in Node 10
It seems that in 2 places in NodeServices we're still using the constructor for Buffer, which was [deprecated](https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/) in node 10. This causes errors like we see [here](https://github.com/aspnet/templating/issues/550). We should do as the exception suggests and use one of `alloc`, `allocUnsafe` or `from`, as appropriate.

An FYI to @PatrickKunz who files aspnet/templating#550.

ryanbrandenburg:
Fixes #1668.

All the whitespace stuff was done by VSCode, but it appears to be correct to me, so I'm not going to fight it unless @SteveSandersonMS objects.

### DIFF
--- a/scripts/Regenerate-JSFiles.ps1
+++ b/scripts/Regenerate-JSFiles.ps1
@@ -1,0 +1,28 @@
+[cmdletbinding(SupportsShouldProcess = $true)]
+param(
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2
+
+Push-Location "src"
+try {
+    $dirs = Get-ChildItem -Directory
+    foreach($dir in $dirs)
+    {
+        Push-Location $dir
+        try{
+            if(Test-Path -Path "package.json")
+            {
+                npm install
+                npm run build
+            }
+        }
+        finally{
+            Pop-Location
+        }
+    }
+}
+finally {
+    Pop-Location
+}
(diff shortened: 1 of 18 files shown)


ryanbrandenburg:
What is `src/Microsoft.AspNetCore.NodeServices.Sockets/Content/Node/entrypoint-socket.js` generated from? How can I re-generate it to ensure it's all correct?

ryanbrandenburg:
I would prefer that we test generated code for functionality rather than check it in to examine diffs. But for now we should at least make it a little clearer what parts of the code are generated.

ryanbrandenburg:
@SteveSandersonMS we need answers to the above questions before we make progress on this work.

SteveSandersonMS:
@ryanbrandenburg The `NodeServices`/`NodeServices.Sockets`/`SpaServices` projects all contain TypeScript code that gets built following typical NPM/Webpack conventions. If you check the `package.json` files in those directories, you'll see they all include `scripts` config entries similar to:

```
  "scripts": {
    "build": "... some command involving webpack ..."
  },
```

This implies you can run the webpack build by using the following commands:

```
cd [that directory containing package json]
npm install
npm run build
```

In the `webpack.config.js` files, each of these projects is configured to place its Webpack build output in `./Content/Node` (see the `output` section of the config). So, that's the subdirectory whose content is generated.

**Note**

The Webpack build configs currently in `master` assume you're using relatively old versions of Webpack and TypeScript. I just checked, and due to changes in Webpack and TypeScript, the Webpack build actually fails if you update to the latest Webpack/TypeScript. Since there's no good reason to require the use of older versions of Webpack/TypeScript, I've just submitted PR #1748 which updates all this build config to the latest Webpack/TypeScript. I recommend we merge this before you attempt to actually use the Webpack builds, otherwise you'll probably get errors.

SteveSandersonMS:
Looks like the unrelated changes are just whitespace, which is fine. Is it possible that you got VS to reformat the files again? If not I’m not sure why the whitespace diff is there. In any case it won’t be a problem.

ryanbrandenburg:
Re(Re)-generating didn't show any difference between the two, so I assume it's just a difference in TypeScript compilation. Will merge once branches are open.

PR #1748, opened 2018-08-28
Update NodeServices/SpaServices/NodeServices.Sockets to build using latest TypeScript and Webpack

SteveSandersonMS:
Previously they assumed you were using versions of TypeScript and Webpack that are by now rather outdated. This PR:

1. Updates to the latest TypeScript/Webpack
2. Includes `package-lock.json` files to lock down exactly which versions should be used, so it won't break in the future if TypeScript/Webpack themselves introduce breaking changes


ryanbrandenburg:
@SteveSandersonMS, this PR targeted master, we need it in release/2.2 as well. I'm going to cherry-pick them back.

SteveSandersonMS:
Do you need it in 2.2 though? It’s just build infrastructure, not changing the resulting product.

SteveSandersonMS:
Well I guess there’s no reason *not* to cherry-pick to 2.2 as well. It will help if we need to make changes in 2.x patches.

PR #1775, opened 2018-10-17
Add optional port parameter for the development servers

stephtr:
As written in #1766 I would find it useful to have the ability to specify a port in `UseReactDevelopmentServer` and `UseAngularCliServer`. With that JavaScriptServices would automatically detect if there is already an active development server running and start a new one otherwise.


johnrom:
This is something I'm interested in for building a usable React SSR! And it doesn't appear to be a breaking change (so far as I can tell).

One thing, I would rather pass a configuration action like most middleware initializers. Then switch the current parameter-based versions to build the options and call the method with the configuration object.

```
public class ReactDevelopmentMiddlewareOptions { 
  public int Port;
  public string NpmScript
}

// options is ReactDevelopmentMiddlewareOptions
app.UseReactDevelopmentServer(options => {
  options.Port = 7878;
});
```

Eilon:

Thanks,
Eilon

PR #1789, opened 2018-11-13
Do not fallback to creating console logger

pakrym:
Falling back to creating instances of console logger is bad practice. Each of them would create a new writing thread and background queue. In addition, ConsoleLogger type is becoming private.

cc @Eilon

Technically it's a breaking change for customers who didn't have logging configured at all and used JavaScriptServices.


Eilon:
Please log an announcement for this. I can't tell how big a breaking change it will be for some people. Please include the new code someone would have to use if they still want to have console logging.

cc @SteveSandersonMS

